PR #5, opened 2016-03-07
Adds filter tile component and fonts

nikhilaravi:
Filter tile component with clickable stars


nikhilaravi:

PR #6, opened 2016-03-07
adds initial tags from dummy data

jackcarlisle:
* Adds tags from mocked data
* Changes background images of the placeholder tiles


nikhilaravi:
niiceee :candy:

PR #8, opened 2016-03-08
Package tile

nikhilaravi:
Adds package tile component. 

https://tc-jira.atlassian.net/browse/ISEARCH-203


jackcarlisle:
looks good! :+1:

PR #13, opened 2016-03-09
Filter tile

nikhilaravi:
Adds functionality for star rating and updates font on the filter tile.


jackcarlisle:
nice :+1:

PR #16, opened 2016-03-09
adds 100% width to image behind search bar

jackcarlisle:
* adds 100% width to search bar image for mobile

TODO
* add media query to support larger devices


nikhilaravi:
:+1:

PR #18, opened 2016-03-09
Grid updates

nikhilaravi:
Removes white border and fixes centering of tiles and stars.

https://tc-jira.atlassian.net/browse/ISEARCH-188


jackcarlisle:
looks good :+1:

PR #21, opened 2016-03-09
Adds info bubble on click of the question mark

nikhilaravi:
Adds info box on filter tile when question mark is clicked. 


jackcarlisle:
nice one :+1:

PR #27, opened 2016-03-10
scroll

jackcarlisle:
* adds scroll to the add message box which takes you to the bottom of the page after being clicked


nikhilaravi:
:+1:

PR #31, opened 2016-03-10
Updated Filter icon

nikhilaravi:
Added floating filter icon with a modal that shows the tags. 
Fixed styling of tags to match the new UI designs


jackcarlisle:
nice :+1:

PR #37, opened 2016-03-17
responsive search summary

jackcarlisle:
* adds  media queries to the search summary bar so that it responds to multiple devices


nikhilaravi:
:100:

PR #43, opened 2016-03-18
Spinner

jackcarlisle:
* adds the loading spinner when waiting for search results to load


jruts:
We need to refactor this later

PR #55, opened 2016-03-23
Component tests

nikhilaravi:
added/updated tests for:
- package tile
- filter tile
- star rating
- tags


lib/filter-tile-yesno/index.js
```diff
@@ -33,7 +34,8 @@ export default class FilterTile extends Component {
               showAddMessage={showAddMessage}
             />
           </div>
-        </div>}
+        </div>
+      }
```
nelsonic: more readable. :+1:

lib/scroll-view/test/index.test.js
```diff
@@ -0,0 +1,26 @@
+// import React from 'react';
+// import { mount } from 'enzyme';
+// import jsdom from 'mocha-jsdom';
+// import ScrollView from '../';
```
nelsonic: guessing we are going to come *back* to this later...?

nelsonic:
looks good. :+1:

lib/scroll-view/test/index.test.js
nikhilaravi: Yes. I need to do some research about how to set the node heights etc.

PR #56, opened 2016-03-24
Refactoring

nikhilaravi:
- refactored containers so now there is only one container (isearch)
- refactored reducers so there is only one reducer (search)
- fixed tests 

- mock tile and tag data is loaded inside the reducer instead of the isearch component


nikhilaravi:
@jackcarlisle @jruts @nelsonic please check and see if the changes make sense!

nelsonic:
@nikhilaravi code looks good. 
Tests pass on CodeShip because the `node_modules` have been cached.
they fail on Travis because, Travis installs from scratch each time and 
cannot find `react-addons-test-utils`
see: https://travis-ci.org/numo-labs/isearch-ui/builds/118212559#L1139

nikhilaravi:
@nelsonic yes I realised that enzyme depends on the react test utils module and you have to install is separately. I have added it back to the dev dependenices

nelsonic:
@nikhilaravi thanks for fixing that. 
As per discussion with @plaenen give that we will be updating the iSearch UI heavily once the Taggable system is operational, we have decided to focus on https://github.com/numo-labs/taggable-ui for the next week and return to this. :+1:

PR #81, opened 2016-04-18
Article view

Kumjami:
First approach.
Work in progress


nelsonic:
@Kumjami don't forget to add tests and ensure that code coverage does not *decrease*. 
if you have any questions, ask! 👍

lib/article-fullpage/index.js
```diff
@@ -0,0 +1,26 @@
+import React, { Component, PropTypes } from 'react';
+
+require('./style.css');
+
+export default class ArticleFullPage extends Component {
+  render () {
+    const { articleContent } = this.props;
+    return (
+        <div className='articleFullPageContainer'>
```
nikhilaravi: just watch out for the indentation here - think there is an extra space for all the elements inside the return

lib/article/index.js
```diff
@@ -5,7 +5,7 @@ require('./style.css');
 export default class Article extends Component {
   render () {
     return (
-        <div className='articleContainer'>
+        <div className='articleContainer' onClick={() => {window.scrollTo(0, 0);this.props.viewArticle(this.props.content)}}>
```
nikhilaravi: what does the window.scrollTo(0,0) do? Also for readability, you can put each prop on a separate line.

lib/nav-header/index.js
```diff
@@ -0,0 +1,22 @@
+import React, { Component, PropTypes } from 'react';
+
+require('./style.css');
+
+export default class NavHeader extends Component {
+  render () {
+    const { articleContent, backToSearch } = this.props;
+    return (
+      <div className='navBarContainer'>
+          <a className='backButton' onClick={() => backToSearch()}><i className="fa fa-angle-left fa-2x" aria-hidden="true"></i></a>
```
nikhilaravi: same here. It would be more readable to spread this out over multiple lines with each prop on a separate line

src/components/isearch/index.js
```diff
@@ -99,9 +120,22 @@ ISearch.propTypes = {
...
+  backToSearch: PropTypes.func,
+  articlePage: PropTypes.bool,
+  articleContent: PropTypes.object,
   addSearchStringTag: PropTypes.func,
   loading: PropTypes.bool,
   error: PropTypes.string
 };
 
+/**
+ *
```
nikhilaravi: I think the SearchBar component is no longer used so you can remove the comment

src/reducers/article.js
```diff
@@ -0,0 +1,19 @@
...
+  articlePage: false,
+  articleContent: {}
+};
+
+export default function article (state = initialState, action) {
+  switch (action.type) {
+    case VIEW_ARTICLE:
+      return {...state, articlePage: true, articleContent: action.content};
+    case VIEW_SEARCH:
+      return {...state, articlePage: false, articleContent: {}};
```
nikhilaravi: you could just return `{...initialState}`  here

lib/article-fullpage/index.js
Kumjami: true, thanks, removed

jackcarlisle:
@Kumjami do the article full page and article components need to be separate? You could combine them into the same folder but still have different files for example you could have -->

── article
         |── index.js
         |── article-full-page.js
         |── article-footer.js
         |── styles.css

You could import everything into index.js

lib/article/index.js
Kumjami: window.scrollTo(0,0) is just to move scroll to top, if not used, the article is displayed, but the scroll is not at the top

making it more readable, thanks

jackcarlisle:
@Kumjami does the footer need to be called 'article-footer'? It could be extracted into a more generic component that could be reused if it isn't really specific to the footer. In this case you could rename the folder to just 'footer' and then import it into 'article'.

src/components/isearch/index.js
Kumjami: removed

lib/nav-header/index.js
Kumjami: done

src/reducers/article.js
Kumjami: done!

jackcarlisle:
looks good 👍

PR #90, opened 2016-04-21
Tags

nikhilaravi:
- updates rendering of tags
- adds a loading spinner to show more results are being added
- adds new results at the beginning of the feed so the user knows results are being added.


nelsonic:
Nice work. ✅

PR #91, opened 2016-04-21
Drop down

jackcarlisle:
* creates the drop down component

TODO:

*hook up the input fields to the state


nikhilaravi:
looks good. 👍

Issue #96, opened 2016-04-25
Distribution of filter and article tiles in the feed


nikhilaravi:
At the moment, the tiles and articles are not evenly spread out through the feed. Check this function is working as desired! 

https://github.com/numo-labs/isearch-ui/blob/master/src/reducers/utils/helpers.js

PR #106, opened 2016-04-25
Filtering

### LINKED ISSUE #104: Query should not be reissued when a tag is repeated.
When a tag is added twice the tags array is filtered for unique ids so the tag is not added but a new query is issued! This should not happen!

nikhilaravi:
Fixes #104 #96 

- need to update the tests


jackcarlisle:
nice one! ✌🏼 🐧

PR #109, opened 2016-04-26
Tests to validate queries and mutations against GraphQL schema

### LINKED ISSUE #94: Add Tests to validate GraphQL queries
This will help a lot in avoiding 'null' values being returned from GraphQL due to queries that don't match the schema. 

Example implementation on numo-ui 
https://github.com/numo-labs/numo-ui/blob/master/src/constants/__tests__/queries-test.js

nikhilaravi:
Fixes #94 

NB - graphql version on lambda-numo-graphql is 0.4 so the same version needs to be used in this repo as otherwise the introspectionQuery has a different fields and introspection fails.


jruts:
Very good @nikhilaravi :clap:

Issue #126, opened 2016-05-12
Add Pre-Commit Hook to Run Semistandard (Lint) and Run Tests


nelsonic:
At present there is no `pre-commit hook` for the front-end UI. 
at the very least we should be doing a `npm run lint` to ensure that code which does not pass `semistandard` *cannot* be committed to GitHub (*which wastes time on CodeShip etc.*)

The pre-commit hook should also run the tests *locally* before allowing the commit.

In the *future* we should also add a pre-commit hook to check for test coverage.
(*at the moment Coverage on `master` is 85% ...*)

nikhilaravi:
Closing this issue as @nelsonic pre-commit hook has now been added.

PR #129, opened 2016-05-12
Add tiles already when no results come back within 3 seconds

jruts:
Also made sure that the current search is stopped when a new search is triggered


nelsonic:
we *must* add pre-commit hook after demo tomorrow... #126 
tests pass.
coverage down.
coverage quest on monday!

PR #130, opened 2016-05-12
Scroll fix

nikhilaravi:
Saves scroll position of feed and restores scroll position when navigating back from the hotel page. 
Sets the content type of uploaded files using the mime-type module - fixes rendering of svgs on s3!!


nelsonic:
Amazing. :-)

PR #132, opened 2016-05-12
Make sure tiles are present if there are any

jruts:


nelsonic:
looks good. come on codship... 🚢

PR #133, opened 2016-05-12
Article fix

nikhilaravi:
Scrolling of page back after clicking 'add tag' is not working completely as desired but I can fix that tomorrow! :)


nelsonic:
@nikhilaravi looks good. 👍 
merging. :shipit:

PR #135, opened 2016-05-12
Hotel tiles fixs

Kumjami:
Hotel tiles texts and dates fixed (conflict with @nikhilaravi changes).
Hotel page last changes requested by @unnderbar added.


lib/hotel-page/index.js
```diff
@@ -129,12 +129,11 @@ class HotelPage extends Component {
               </div>
               {this.renderFactlist(this.retrieveAmenities())}
             </div>
-            <div className='bookButtonContainer'>
+            <div className='bookButtonContainer clearfix'>
               <div className='ppp'>ppp <span className='hotelPrice'>
-                {packageOffer.price.perPerson},-
+                {parseInt(packageOffer.price.perPerson, 10).toLocaleString('da-DK')},-
```
nelsonic: do we have a *test* for this?

Kumjami: good point, updating

Kumjami: It's not working at expected at 'server-side' while testing. 
So... should I remove it? The idea behind that was to be able to select based on locale (right now hard-coded to `da-DK` the proper number format (requested by @unnderbar). 
What I would do for sure is changing `parseInt` to `parseFloat` to be able to parse something like `12345.67` into `12.345,67`, with `parseInt` it would be `12.345`. 
So, the question. Should I remove that?. What's your opinion?, @nelsonic

nelsonic: Please add an issue to the backlog to create a test for this. 
merging now to get this in.

PR #147, opened 2016-05-18
[WIP] reviewing components

Kumjami:
Removing unused code and refactoring components.


nikhilaravi:
👍 I think we still need the `viewArticle` and `viewHotel` actions to actually fetch the data so I'm going to add them back in!

PR #150, opened 2016-05-19
History fix

nikhilaravi:
- added configuration for browserHistory but reverted back to hashHistory so the website is functional. 

- will investigate cloudfront routing in more detail and then try and reimplement browserHistory

### DIFF
--- a/config.js
+++ b/config.js
@@ -1,0 +1,1 @@
+export const basename = '/isearch/0.14/index.html';
--- a/README.md
+++ b/README.md
@@ -45,20 +45,6 @@ To move a component out into its own npm module
 * Copy over the contents of the component from the lib folder
 * Create a font folder and copy over the necessary font files
 
-## Deployment to S3
-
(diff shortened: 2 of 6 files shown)


config.js
```diff
@@ -0,0 +1 @@
+export const basename = '/isearch/0.14/index.html';
```
nelsonic: could you not derive this by doing `require('pacakge.json').version.split('.')[2]` to avoid having to change it on every release...?

nikhilaravi: Ah yes could do that. I have added a line in the gulpfile which does something similar but your suggestion might be simpler!

nelsonic:
Travis is taking his sweet time... all those Yanks are online blitzing the poor Travis servers...
(*Code looks good tho...*)

config.js
nelsonic: issue created: https://github.com/numo-labs/isearch-ui/issues/154 resolve on next release. 👍

PR #151, opened 2016-05-19
Websockets

nikhilaravi:
implements web socket service to connect to the web socket server


src/index.html
```diff
@@ -46,6 +46,12 @@
     <!-- <script src="https://cdn.rawgit.com/robflaherty/riveted/master/riveted.min.js"></script>
     <script>riveted.init();</script> -->
     <!-- End Google Tag Manager -->
+    <!-- Web socket connection -->
+    <script src="http://eb-ci.wmm63vqska.eu-west-1.elasticbeanstalk.com/primus/primus.js"></script>
```
nelsonic: I was wondering if there was any way for us to "obfuscate" this URL to avoid it getting DDOSed ... but *realistically* there isn't a *simple* way...

src/index.html
```diff
@@ -46,6 +46,12 @@
...
+    <script>
+      primus = new Primus('http://eb-ci.wmm63vqska.eu-west-1.elasticbeanstalk.com');
```
nelsonic: we should defs have a variable for for the URL so it is not repeated.

src/services/websockets.js
```diff
@@ -0,0 +1,14 @@
+import * as SearchResultActions from '../actions/search-results.js';
+
+export function initialise (actionCreatorBinder) {
+  const {
+    saveSearchResult
+  } = actionCreatorBinder(SearchResultActions);
+
+  primus.on('data', function received (data) {
```
nelsonic: looks pretty standard so far. 👍

nelsonic:
Looks good so far. 👍

nelsonic:
@nikhilaravi given that this was working on your localhost on Friday and the tests have passed, I'm removing the `[WIP]` Label and reviewing. :

nikhilaravi:
@nelsonic just adding a few updates - the tiles are being retrieved through the socket channel, just need the packages now to test it's all working!

nikhilaravi:
@nelsonic I've removed the primus script tags and added it in as a file instead. There are tests for the web socket service which pass when I run them separately but fail when run with the other tests!! Can't figure out whats wrong... If you or @tomgco could have a look that would be great! It's in the file `test/services/websockets.test.js`

PR READY TO MERGE!

nelsonic:
@nikhilaravi I had the same problem when trying to run _real_ tests against Primus.
The only way I could get the entire suite of tests to run (_including WebSocket tests_) 
was to rename the file so it gets run *first*:
https://github.com/numo-labs/aws-lambda-helper/blob/master/test/lib/_pushToSocketServer.test.js#L1

nikhilaravi:
@nelsonic that's annoying but glad to hear it wasn't just me!

nelsonic:
@nikhilaravi annoying *indeed*. maybe @tomgco can enlighten us... ❓

lennym:
The issue is that the primus instance defined here - https://github.com/numo-labs/isearch-ui/pull/151/files#diff-e5dc273ff838fcfd0afb925ff66c6e53R13 is a common instance across all modules, so connects once the first time it is required, but then does not need to reconnect.

That means that the websocket test does not receive a 'connected' event (since this was emitted somewhere earlier on in the test suite the first time the websocket service module was required) and so does not post data, and so in turn does not receive any subsequent events.

Instead the test wait for a connection event that occurred some time previously until it times out.
There's a number of options for a solution:

* refactor the websocket service so that the primus instance is created in a function call rather than at require time. This means that different implementations will always use separate primus instances, and so will always emit a connection event.
* tear down the socket connection after each test so that each new test always connects
* establish if the socket is already connected in the websocket test, and perform the `postDataToClient` immediately if a connection is present.

src/index.html
tomgco: This can be put into S3 at some point, the only requirement is that it matches the version of primus that is currently deployed. This endpoint is purely for development.

lennym:
I raised https://github.com/numo-labs/isearch-ui/pull/160/files which effectively performs Option 1 I gave above, and fixes the issue. I wouldn't recommend renaming the file as a solution as it can work exactly once - only one test can execute first, as soon as we add a second websocket test it will fail again.

It won't merge right now on account of the file name change, but if that could be reverted then it should be good to go.

Kumjami:
Agree with @lennym

src/actions/search-results.js
```diff
@@ -278,59 +115,42 @@ function constructDepartureAirportQuery (departureAirport) {
...
-      const formattedTags = formatQuery(tags);
-      const passengers = combinePassengersForQuery(childAgeArray, numberOfChildren, numberOfAdults);
-      const departureAirports = constructDepartureAirportQuery(departureAirport);
-      const travelPeriod = constructTravelPeriodQuery(departureDate, duration);
-      const query = {passengers: passengers, travelPeriod: travelPeriod, departureAirports: departureAirports, ...formattedTags};
+      const query = formatQuery(store);
       console.log('query', JSON.stringify(query));
       return graphqlService
-        .query(MUTATION_START_SEARCH, {'query': JSON.stringify(query)})
+        .query(MUTATION_START_SEARCH, {'query': JSON.stringify(query), clientId, connectionId})
```
nelsonic: we *also* need to know the `userId` of the client. this is the `id` that is saved on the device and will be re-used if the person has multiple sessions.

nelsonic:
looks like this is ready. (*tests passing*)

PR #155, opened 2016-05-20
Article tag fix

### LINKED ISSUE #149: "New search" bug
While trying to create a new search bucket, there is a bug if the graphql mutation params include `marketing` and `tile` filters type. 

Using the following query: 
```js
mutation startSearch($query: String) {
  viewer {
    searchResultId(query: $query) {
      id
    }
  }
}
```
And the following query variables: 
```js
{
 "query": "{\"passengers\":[{\"birthday\":\"1996-04-18\"},{\"birthday\":\"1996-04-18\"}],\"travelPeriod\":{\"departureBetween\":[\"2016-06-01\"],\"nights\":[7]},\"departureAirports\":[\"airport:master.CPH\"],\"amenity\":[\"amenity:ne.wifi\"],\"marketing\":[\"marketing:homepage.dk.spies\"],\"tile\":[\"tile:article.dk.62\"]}" 
}
```

We will get: 
```js
{
  "data": {
    "viewer": {
      "searchResultId": {
        "id": null
      }
    }
  }
}
```

If there are `tile` or `marketing` filter, but not both, the id would be a valid UUID.

nikhilaravi:
fixes #149 
Adds tests!

### DIFF
--- a/lib/package-tile/index.js
+++ b/lib/package-tile/index.js
@@ -67,6 +67,5 @@ export default class PackageTile extends Component {
 PackageTile.propTypes = {
   packageOffer: PropTypes.object,
   totalPassengers: PropTypes.number,
-  itemId: PropTypes.string,
-  bucketId: PropTypes.string
+  itemId: PropTypes.string
 };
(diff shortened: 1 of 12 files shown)


jruts:
:'( codecov

PR #160, opened 2016-05-25
Return primus instance from websocket service initialise method

lennym:
Previously all uses of the websocket service used a shared primus instance which meant that the state of the websocket would be in an unpredictable state when running unit tests depending on the ordering of the tests.

Instead create a new primus instance per call to `initialise`, which means that the socket will always be in a fresh known state when running tests against it.

### DIFF
--- a/src/services/websockets.js
+++ b/src/services/websockets.js
@@ -10,9 +10,9 @@ const socketUrl = 'http://eb-ci.wmm63vqska.eu-west-1.elasticbeanstalk.com';
 * @param {Function} - actionCreatorBinder - function that takes an action
 * and binds it to dispatch
 */
-export const primus = new Primus(socketUrl);
 
 export function initialise (actionCreatorBinder) {
+  const primus = new Primus(socketUrl);
   const {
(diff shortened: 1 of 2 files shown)


nelsonic:
@lennym this looks good. if you can resolve merge conflict(s) we could merge this in. 👍

Kumjami:
@lennym can you please fix the conflicts so I could merge it?

nikhilaravi:
@lennym great!

lennym:
I've rebased to remove the conflicts.

However, I'm not sure if now we might want to add a teardown for the socket to `componentWillUnmount` to correspond to https://github.com/numo-labs/isearch-ui/blob/websockets/src/containers/router.js#L32 to prevent multiple sockets being left open if the component is mounted multiple times. My knowledge of React is a little too embryonic to know for certain.

Kumjami:
This should be triggered once. But I think it could be a good idea to have a teardown, as a default procedure: initialize and remove. 
I'm merging this to `websockets`. 
:+1:

PR #161, opened 2016-05-25
Scroll view

nikhilaravi:
- Adds ScrollView component around the grid - this loads 5 tiles into the feed each time the user scrolls to the bottom of the page. 

- Need to fix the addition of filters into the results - the RECIEVE_SEARCH_RESULTS action in the search reducer - @jruts can you take a look at this?

Only merge this PR after the web sockets PR has been merged!


nelsonic:
@nikhilaravi please confirm when this is ready for review. (_we're excited to see it! thanks_!)

nikhilaravi:
@nelsonic I'm going to try and write some tests for the scroll behaviour in React - as its going to be such a crucial part of the app want to make sure it works as expected..! Just discovered one bug the hard way!

src/components/search-results/index.js
```diff
@@ -7,13 +7,17 @@ import VisbilitySensor from 'react-visibility-sensor';
...
+  transitionDuration: '0.2s',
   itemSelector: '.gridItem',
   fitWidth: true,
   gutter: 14 // horizontal spacing between tiles
 };
 
 class SearchResults extends Component {
+  constructor () {
+    super();
+    this.mapItems = this.mapItems.bind(this);
```
nelsonic: ❓ (*sorry, I don't understand why we need this...*)

nelsonic:
@nikhilaravi code looks good. might benefit from the *occasional* in-line comment to ensure "*future dev*" understands what's going on. but happy to merge.

src/components/search-results/index.js
nikhilaravi: The mapItems function needs to access `this.props` and with React Classes you need to explicitly bind `this` to the Class methods.

nikhilaravi:
@nelsonic okay sure will add more comments in future!

PR #166, opened 2016-05-25
Reduce binary size from 2MB to ~800KB

tomgco:


lennym:
"...with this one weird trick"

nelsonic:
@lennym are you investigating why the build is failing on `master` ?

lennym:
It looks to be because the websocket server (a.k.a numo-labs/graphql-backchannel) was playing up. Builds are green now.

lennym:
@tomgco You score a 👎  for starting a branch name with a `m` and so breaking my console tab completion of `master`.

😠

PR #167, opened 2016-05-25
Teardown socket instance if router container unmounts

### LINKED ISSUE #162: Teardown websocket instance.
As @lennym [propose](https://github.com/numo-labs/isearch-ui/pull/160#issuecomment-221575681). It could be nice to have a teardown for websocket at `componentWillUnmount` to be sure that it's closed if the element is unmounted somehow.

lennym:
This will prevent any sockets potentially being left open or created if the router component is mounted multiple times.

Fixes #162


nelsonic:
@lennym looks good. 👍

Issue #118, opened 2016-05-05
UI fixes


jackcarlisle:
TODO: 

* [x] **Hotel page:** add hotel description to the hotel page ‼️
* [x] **Searchbar:** remove hotels from the dropdown searchbar ❗️
* [x] change the 'book' button on the article page to an 'add tag' button 
* [x] **Inspiration Feed:**  include discount  price if there is one (Discount = pink price, regular price = grey price) ❗

* [x] replace price per person with the total booking price 

* [x] add home page default query tag - Denmark related - gets removed when a user performs an action to add a tag themselves
* [x] switch the kids and adults dropdowns around in the travel info section
* [x] add another button at the bottom of the hotel page that links back to spies.dk
* [x] move the content on the article and hotel pages up into the image at the top of the page
* [x] remove media query that changes the width of the grid tiles
* [x] increase vertical and horizontal spacing of the grid to 20px

nikhilaravi:
More UI Fixes

**Hotel page**
* [ ] Remove the slide for the images on desktop and display them as a list instead
* [ ] centre the amenities text and make sure the book box doesn't overlap the description text

**Package Tile**
* [ ] Change the english words to Danish (nights, months etc)
* [ ] Check why the star rating is not showing up for some packages

** Home Page**
* [ ] Hero image with the search bar at the bottom - the page should scroll upwards when the input box is clicked (so some tiles are visible and the autocomplete box shows up)
* [ ] Add a load more button at the end of the feed to load more content
* [ ] Fade in the images after they are fully loaded
* [ ] search bar should take up more width in desktop and mobile - increase height and width and add margin between the search bar and tags
*  [ ] Add signpost svg in the header image next to the title text

**Nav Bar**
* [ ] Should show the duration of the trip in Danish 
*  [ ] the edit button should be a pen (not a / ) (icon in the UX folder)

Stretch goals!
* [ ] bouncy arrow on the home page to indicate the user to scroll down
* [ ] parallax scroll of the header image
* [ ] lift and shadow on hover over each tile - just for deskptop

nelsonic:
Can this issue be closed...?

nikhilaravi:
@nelsonic I've moved over the UI enhancements to a new issues so yes this can be closed!

PR #170, opened 2016-05-27
Remove tile

jackcarlisle:
## ready for review

#118 
* adds functionality to remove a tile from the list of grid items


nelsonic:
@jackcarlisle functionality *looks* good, please add a test so coverage does not decrease. thanks! 👍

jackcarlisle:
@nelsonic will do!

src/assets/cancel.svg
```diff
@@ -0,0 +1,41 @@
...
+<g>
+</g>
+<g>
+</g>
+<g>
+</g>
+<g>
+</g>
+<g>
+</g>
```
lennym: Probably worth removing these `<g></g>`.

jackcarlisle:
@lennym thanks will do!

lennym:
My only thought is that this will likely cause some slightly odd behaviour once infinite scrolling is also brought in since I think it will it will mean that some search results won't be displayed if others are removed.

Example:

1) user loads results 0-9
2) user removes a result
3) user scrolls, triggering rendering of a slice of results starting from index 10 (see [code](https://github.com/numo-labs/isearch-ui/blob/24f0e25d271ad6866d5fea7fdaedfacd2b14df1b/src/components/isearch/index.js#L38))
4) since result 10 is now at index 9 on account of filtering, it is never rendered.

I'm not really sure if this is a problem.

nelsonic:
@lennym good observation. would it be better to have a `visible` property for each tile and simply flip it to false instead of removing the item from the `state` ?

lennym:
That makes sense as a solution, though I'd probably use a `hidden` flag rather than `visible` so we can default to visible without having to set the property everywhere.

jackcarlisle:
@lennym good point!
@nelsonic I'll add the hidden property

nelsonic:
@jackcarlisle is this good to go now then...?

jackcarlisle:
@nelsonic yep!

nelsonic:
@jackcarlisle sweet! 🍭

PR #172, opened 2016-05-27
Undefined region fix

jackcarlisle:
* adds fix for case where the region is undefined in the package offer


nelsonic:
@jackcarlisle "_makes sense_" ...

PR #173, opened 2016-05-27
Use production for production deploys

tomgco:


webpack.production.config.js
```diff
@@ -38,6 +38,11 @@ module.exports = {
     ]
   },
   plugins: [
+    new webpack.DefinePlugin({
+      'process.env': {
+        'NODE_ENV': JSON.stringify('production')
```
lennym: What's the logic behind `JSON.stringify`-ing a string?

tomgco: Was copied and pasted from the webpack website. It is because the string needs to be like this:

`'NODE_ENV': '"production"'`

lennym: Euww... that's N.A.S.T.Y.

Fair enough though.

PR #174, opened 2016-05-31
Fix formatting of default birthdays

lennym:
The value returned by `Date#getMonth` is zero-indexed (see [docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getMonth)) so if we use the value to construct a YYYY-MM-DD date string then we need to add one to it to ensure a valid date is produced.

Currently date validation is failing because the dates returned are the 31st April.

### DIFF
--- a/src/actions/helpers.js
+++ b/src/actions/helpers.js
@@ -67,14 +67,14 @@ export function combinePassengersForQuery (childAgeArray, numberOfChildren, numb
   const childPassengers = slicedChildAgeArray.map(slicedChildAge => {
     const date = new Date();
     const year = date.getFullYear() - Number(slicedChildAge.split(' ')[0]);
-    const month = ('0' + date.getMonth()).slice(-2);
+    const month = ('0' + (date.getMonth() + 1)).slice(-2);
     const day = ('0' + date.getDate()).slice(-2);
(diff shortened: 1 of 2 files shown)


nelsonic:
@lennym code looks good. might need to increase test timeout to ensure tests pass.
I'm sure you're already on the case...

nelsonic:
@lennym bizarrely the websocket test is still failing. 
this branch *appears* to only have *one commit* on it, did you *squash* your commits?
if you haven't already, please merge master into the `bugfix/date-formatting` branch to get the change merged in PR https://github.com/numo-labs/isearch-ui/pull/175 so the tests will pass for *this* PR.

lennym:
@nelsonic It's not bizarre, the websockets tests were fully broken. There's a fix in #179

nelsonic:
@lennym thanks for diving into it and fixing. 👍

nelsonic:
@lennym PR #179 has been merged. (*thanks again BTW!*)
Ping me when you've updated this branch/PR to include the fix required to make this pass. 👍

lennym:
@nelsonic Looks like it's finally good to go 😄

nelsonic:
Cool. thanks! 🎉

PR #179, opened 2016-05-31
Refactor websockets tests to not depend on remote service

lennym:
The previous version of the tests was sending POST requests to a remote service to trigger returning websockets events. This is no longer a supported interface, so stopped working.

Instead we can simplify the triggering of events by calling `emit` directly on the primus instance, which has no dependency on others services being available.


nelsonic:
@lennym thanks for fixing this. 👍

PR #181, opened 2016-05-31
Hide tile

jackcarlisle:
## ready for review

* adds functionality to hide a tile but not removed from the state when 'x' is clicked


jackcarlisle:
screen shot 2016-05-31 at 15 05 38

When using the hidden method, the tiles leave gaps when they are hidden after the results have all been loaded. This _(hopefully)_ shouldn't be an issue when we implement infinite scroll.

lib/package-tile/index.js
```diff
@@ -30,35 +36,50 @@ export default class PackageTile extends Component {
     const totalPriceBeforeDiscount = oldPricePerPerson * totalPassengers;
     return (discountPrice && discountPrice !== '0') ? (totalPriceBeforeDiscount.toLocaleString('da-DK') + ',-') : '';
   }
+  handleOnClick () {
+    this.setState({
+      visible: false
```
nelsonic: @jackcarlisle if you instead use `display:none` you will not see the blank spaces on the page identified in https://github.com/numo-labs/isearch-ui/pull/181#issuecomment-222702851

> see: http://stackoverflow.com/questions/133051/what-is-the-difference-between-visibilityhidden-and-displaynone

jackcarlisle:
@nelsonic ok cool I'll do that instead!

nelsonic:
@jackcarlisle sounds like we can simply *remove* the item from the list. its cleanererer. 👍

jackcarlisle:
@nelsonic cool, I'll close this PR

nelsonic:
@jackcarlisle all you need to do is go back a couple of commits...

PR #182, opened 2016-05-31
Refactor websocket connection events to persist room id

### LINKED ISSUE #180: UI re-initialises search/tags when websocket reconnects
When the websocket connection disconnects and reconnects the search parameters are re-initialised and the "Top inspiration" tag is displayed alongside any other tags that were already in place. A new search is carried out withe new parameters.

## Steps to reproduce

* Select a search tag and conduct a search
* Force a socket reconnect - I disabled and re-enabled wifi (Note: using Chrome devtools network condition simulator does not disconnect websockets)

## Expected:

The current search is persisted, with only previously selected tags shown.

## Actual

The "Top inspiration" tag is added and a new search is commenced.

lennym:
When the websocket reconnects following a network loss it was reconnecting to a new room id, which prevents it from being able to receive search results from any in flight searches. Instead save the room id only once, and when the socket reconnects, re-subscribe to the same room id.

Additionally, ensure that the default tags are set only on the first connection and not on any subsequent reconnections.

Fixes #180

### DIFF
--- a/src/services/websockets.js
+++ b/src/services/websockets.js
@@ -1,7 +1,7 @@
 import * as SearchResultActions from '../actions/search-results.js';
 import * as TagActions from '../actions/tags.js';
 import Primus from '../../src/services/primus.js';
-const socketUrl = 'http://eb-ci.wmm63vqska.eu-west-1.elasticbeanstalk.com';
+const socketUrl = 'http://eb-ci.wmm63vqska.eu-west-1.elasticbeanstalk.com?auto_room=false';
 /**
 * Function that initialises a connection with the web socket server and saves
 * the id to the redux store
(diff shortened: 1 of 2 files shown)


src/services/websockets.js
```diff
@@ -1,7 +1,7 @@
 import * as SearchResultActions from '../actions/search-results.js';
 import * as TagActions from '../actions/tags.js';
 import Primus from '../../src/services/primus.js';
-const socketUrl = 'http://eb-ci.wmm63vqska.eu-west-1.elasticbeanstalk.com';
+const socketUrl = 'http://eb-ci.wmm63vqska.eu-west-1.elasticbeanstalk.com?auto_room=false';
```
nelsonic: @jruts does this work with `Tracy`...?

src/services/websockets.js
```diff
@@ -20,26 +20,32 @@ export function initialise (actionCreatorBinder) {
...
   primus.on('data', function received (data) {
     console.log('incoming socket data', data);
-    if (data.connection) {
-      saveSocketConnectionId(data.connection);
+    if (data.graphql) {
+      saveSearchResult(data);
+    }
+  });
+
+  primus.once('open', (data) => {
```
nelsonic: @lennym does this still support the scenario you described in #180 where your device looses Wifi/3G ? (*will it re-connect*)

src/services/websockets.js
lennym: @nelsonic @jruts Not sure what a `Tracy` is, or what impact it might have. THis is taking advantage of [this code](https://github.com/numo-labs/graphql-backchannel/blob/master/app.js#L50-L51) in backchannel.

src/services/websockets.js
lennym: Yes, the re-connection is automatic, and triggers the `reconnected` event at https://github.com/numo-labs/isearch-ui/pull/182/files#diff-e5dc273ff838fcfd0afb925ff66c6e53R32 which re-subscribes to the pre-existing primus room.

lennym: For what it's worth, this is one of the things that primuc [claim to be a major selling point of their module](https://github.com/primus/primus#advantages)

nelsonic: @lennym yes, and on reflection, this does not *look* like it will affect `Tracy` because you are still `join`ing the room just making sure it's only done `once`. 👍

nelsonic:
@lennym looks good. 💄 
@jruts please confirm tomorrow that everything still works on your side. 👍

Issue #183, opened 2016-05-31
edit tags, "cancel" (X) and down arrow Icons not showing


nelsonic:
![numo-ui-icons-not-showing](https://cloud.githubusercontent.com/assets/194400/15683797/0b95ac06-275c-11e6-9308-9f909f4cc0a9.png)
@jackcarlisle can you help investigate/resolve this?

nelsonic:
Also, the "Cancel" (X) button...
![cancel-button-missing](https://cloud.githubusercontent.com/assets/194400/15714155/53d038c8-2811-11e6-8e89-d7e2b5310b49.png)

nelsonic:
I can *confirm* that these icons are now showing correctly:
![ui-icons-appear-fixed](https://cloud.githubusercontent.com/assets/194400/15716601/db9fbf36-281a-11e6-9730-1c313050939b.png)

Thanks for fixing it @jackcarlisle 👍

PR #184, opened 2016-06-01
Travel info routing

jackcarlisle:
## ready for review 

#183 

* adds routing to the travel info section 
* fixes favicon and icons


lib/article-tile/index.js
```diff
@@ -3,11 +3,13 @@ import React, { Component, PropTypes } from 'react';
 require('./style.css');
 
 export class ArticleTile extends Component {
-  shouldComponentUpdate (nextProps) {
-    if (nextProps.tile.name === this.props.tile.name) {
-      return false;
-    }
-  }
+  // shouldComponentUpdate (nextProps) {
```
nelsonic: prefer code to be removed instead of commented out unless you are testing something in which case leave an explanatory comment above the commented out code...

lib/hero-image-header/index.js
```diff
@@ -1,6 +1,8 @@
 import React, { Component, PropTypes } from 'react';
 require('./style.css');
 const signpost = require('../../src/assets/goingplaces.svg');
+const bouncingArrow = require('../../src/assets/down-arrow.svg');
```
nelsonic: didn't realise you *could* `require` an `.svg` file... presumably webpack enables this *magic*? 🎩

lib/package-tile/index.js
```diff
@@ -9,11 +9,13 @@ require('./style.css');
 
 export default class PackageTile extends Component {
 
-  shouldComponentUpdate (nextProps) {
-    if (nextProps.packageOffer.hotel.name === this.props.packageOffer.hotel.name) {
-      return false;
-    }
-  }
+  // shouldComponentUpdate (nextProps) {
```
nelsonic: same.

lib/scroll-view/test/index.test.js
```diff
@@ -6,7 +6,7 @@ import expect from 'expect';
 import sinon from 'sinon';
 
 describe('<ScrollView /> Test', () => {
-  jsdom();
+  jsdom({skipWindowCheck: true});
```
nelsonic: https://github.com/rstacruz/mocha-jsdom#special-config

src/components/search-results/index.js
```diff
@@ -111,9 +112,9 @@ class SearchResults extends Component {
               <VisbilitySensor key={index} onChange={(isVisible) => this.handleVisibility(isVisible, item)}>
                 <div key={index} className='gridItem'>
                   <div onClick={() => removeTile(item.id)}>
-                    <img className='removeTileButton' src='../../src/assets/cancel.svg' alt='cancel' />
+                    <img className='removeTileButton' src={removeTileButton} alt='cancel' />
                   </div>
-                  <div onClick={() => changeRoute(`/article/${item.url}`)}>
+                  <div onClick={() => changeRoute(`/article/${resultId}/${item.id}`)}>
```
nelsonic: using the `url` sent in the package/article means we can change it later and not have to update the UI.
this will still work. however it may conflict with what @Kumjami is trying to debug...

lib/hero-image-header/index.js
jackcarlisle: yeah webpack requires that you require the svg in first and then assign it to the image src because it bundles the assets folder into the font folder so if you explicitly say that you need for example an svg from ../../src/assets/cancel.svg it won't be able to resolve the file path

lib/article-tile/index.js
jackcarlisle: removing now

src/index.template.html
```diff
@@ -3,6 +3,7 @@
   <head>
     <link href='https://fonts.googleapis.com/css?family=Source+Sans+Pro:400,400italic,600,600italic,700,700italic,900,300,300italic,200,200italic,900italic' rel='stylesheet' type='text/css'>
     <meta charset="utf-8">
+    <link rel="shortcut icon" href="http://www.spies.dk/StaticContent/web/common/img/favicon-00.ico">
```
nelsonic: thanks for adding that. 👍

nelsonic:
@jackcarlisle ping me when you've made changes. 👍

lib/scroll-view/test/index.test.js
jackcarlisle: it was added as true so that mocha jsdom doesn't complain if jsdom is called in multiple tests

src/components/search-results/index.js
```diff
@@ -75,9 +76,9 @@ class SearchResults extends Component {
             <VisbilitySensor key={index} onChange={(isVisible) => this.handleVisibility(isVisible, item)}>
               <div className='gridItem'>
                 <div onClick={() => removeTile(item.id)}>
-                  <img className='removeTileButton' src='../../src/assets/cancel.svg' alt='cancelled' />
+                  <img className='removeTileButton' src={removeTileButton} alt='cancelled' />
                 </div>
-                <div key={index} onClick={() => changeRoute(`/hotel/${item.url}`)}>
+                <div key={index} onClick={() => changeRoute(`/hotel/${resultId}/`)}>
```
nelsonic: oops. this will no longer work... it should be `item.url` ... @Kumjami please confirm.

nelsonic:
cool. looks good. ✅ 
please avoid using the same commit message multiple times. *ideally* be specific on each commit. e.g: "removes commented code in /component/index.js as discussed in #xyz" 👍

jackcarlisle:
@nelsonic ok great, will do that in future!

PR #187, opened 2016-06-01
TagView Routing

jackcarlisle:
## ready for review

* adds routing to the tagview component in the search summary nav header

### DIFF
--- a/lib/search-summary/tag-view/index.js
+++ b/lib/search-summary/tag-view/index.js
@@ -1,5 +1,6 @@
 import React, { Component, PropTypes } from 'react';
 const closeImage = require('../../../src/assets/close.svg');
+import { Link } from 'react-router';
 
 export default class TagView extends Component {
   render () {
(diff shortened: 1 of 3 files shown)


nelsonic:
@jackcarlisle ideally we should have some sort of test that confirms this is working 
(*and prevents future code from breaking it...*)

jackcarlisle:
@nelsonic ok cool I'll add some tests

lib/search-summary/tag-view/index.js
```diff
@@ -1,18 +1,21 @@
...
     return (
       <div className='blueContainer'>
         <div className={'tagViewActive dropDown'}>
-          <img
-            src={closeImage}
-            alt='exit button'
-            className='tagExitButton'
-            onClick={exitButtonClick}
-          />
+          <Link to='/'>
```
nikhilaravi: can you change this to use a redux action instead of hard coding a Link tag?

nelsonic:
@jackcarlisle let's try and get this finished *first* thing on Monday morning. 🕣  👍

lib/search-summary/index.js
```diff
@@ -65,24 +57,6 @@ export default class SearchSummary extends Component {
...
-      );
-    } else if (visibleTab === 'tags') {
-      return (
-        <div className='searchSummaryContainer'>
-          <TagView exitButtonClick={this.closeTab}/>
-        </div>
-      );
     } else {
       return <div/>;
     }
```
nikhilaravi: Can you clean up this component? There are lots of proptypes which are no longer being used.  The `visibleTab` state can also be change so it is just `visible: true` or `visible: false`.

PR #190, opened 2016-06-02
hotel rendering bug

Kumjami:
fix bug that cause not to render packages while some amenities are boolean instead of strings

### DIFF
--- a/src/components/hotel/index.js
+++ b/src/components/hotel/index.js
@@ -30,7 +30,7 @@ class HotelPage extends Component {
     for (var key in amenitiesObject) {
       if (!amenitiesObject.hasOwnProperty(key)) continue;
       if (key === fact) {
-        if (amenitiesObject[key] === typeof boolean || amenitiesObject[key] === 'true' || amenitiesObject[key] === 'false') {
+        if (typeof amenitiesObject[key] === 'boolean' || amenitiesObject[key] === 'true' || amenitiesObject[key] === 'false') {
           return '-';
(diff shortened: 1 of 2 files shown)


src/components/hotel/index.js
```diff
@@ -30,7 +30,7 @@ class HotelPage extends Component {
     for (var key in amenitiesObject) {
       if (!amenitiesObject.hasOwnProperty(key)) continue;
       if (key === fact) {
-        if (amenitiesObject[key] === typeof boolean || amenitiesObject[key] === 'true' || amenitiesObject[key] === 'false') {
+        if (typeof amenitiesObject[key] === 'boolean' || amenitiesObject[key] === 'true' || amenitiesObject[key] === 'false') {
```
lennym: Given that all the other conditions assume the value type to be `string`, would it be safer to check `typeof amenitiesObject[key] !== 'string'`?

Kumjami: Then 'true' or 'false' would be evaluated as a string, not as boolean.

lennym: I meant in addition to the `=== 'true'` and `=== 'false'` clauses, not in place of the entire condition. So the specific `'true'` and `'false'` values would still be matched.

i.e. 

```javascript
if (typeof amenitiesObject[key] !== 'string' || amenitiesObject[key] === 'true' || amenitiesObject[key] === 'false') {
...
}
```

nelsonic:
@Kumjami please assign this PR for review when ready. 👍

nelsonic:
merging. 👍

Issue #189, opened 2016-06-02
Scrolling restores previously removed tiles


lennym:
If I remove a panel from the near the top of the search results, and then scroll down to load more items, when I scroll back to the top the previously removed item has been restored.

### Steps to reproduce

1) wait for search results to load
2) click the remove button on the top result
3) scroll down far enough to load more scroll items
4) scroll back to the top of the search results

### Expected result

Removed panel is still not visible

### Actual result

Removed panel is restored to view

PR #191, opened 2016-06-02
Bouncing arrow scroll

jackcarlisle:
## ready for review

* adds smooth scroll to the bouncing arrow

### DIFF
--- a/lib/hero-image-header/index.js
+++ b/lib/hero-image-header/index.js
@@ -7,7 +7,6 @@ const bouncingArrow = require('../../src/assets/down-arrow.svg');
 import SearchBar from '../search-bar';
 export default class Header extends Component {
   handleOnArrowClick () {
-    console.log('hello');
     scroll.scrollTo(window.innerHeight / 3);
   }
   render () {
--- a/lib/hero-image-header/index.js
+++ b/lib/hero-image-header/index.js
@@ -6,8 +6,14 @@ const signpost = require('../../src/assets/goingplaces.svg');
 const bouncingArrow = require('../../src/assets/down-arrow.svg');
 import SearchBar from '../search-bar';
 export default class Header extends Component {
+  constructor () {
+    super();
+    this.state = {
+      gridItemsDisplayed: true
+    };
+  }
   handleOnArrowClick () {
-    scroll.scrollTo(window.innerHeight / 3);
+    scroll.scrollTo(window.innerHeight / 1.05);
   }
   render () {
     const {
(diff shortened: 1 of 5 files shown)


lib/hero-image-header/index.js
```diff
@@ -1,10 +1,20 @@
 import React, { Component, PropTypes } from 'react';
+var Scroll = require('react-scroll');
```
nelsonic: I would *suggest* adding some sort of detail in an issue or readme as to why we chose to use https://github.com/fisshy/react-scroll (*despite it's lack of tests...*), are we relying on the fact that it has a few thousand users to file issues and debug it if there is a browser incompatibility...? It works *now*...

lib/hero-image-header/index.js
```diff
@@ -1,10 +1,20 @@
...
 import SearchBar from '../search-bar';
 export default class Header extends Component {
+  constructor () {
+    super();
+    this.state = {
+      gridItemsDisplayed: true
+    };
+  }
+  handleOnArrowClick () {
+    scroll.scrollTo(window.innerHeight / 1.05);
```
nelsonic: this *should* work reliably. 👍

nelsonic:
Code looks good. 
Biggest change is in the schema. (`test/constants/helpers/schema.json`)
*Do* we want to add a test for the functionality before merge...?

jackcarlisle:
Yeah I'll add some tests now!

lib/hero-image-header/index.js
nikhilaravi: I guess we should have checked that - basically we needed a way of scrolling smoothly. We can look for a different module if necessary.

lib/hero-image-header/test/index.test.js
```diff
@@ -6,7 +6,7 @@ import Header from '../';
 describe('Component', function () {
   describe('<Header />', function () {
     it('Should render Header component', function (done) {
-      const wrapper = shallow(<Header />);
+      const wrapper = shallow(<Header displayedItems={[]}/>);
       expect(wrapper.find('.headerBarWrapper')).to.have.length(1);
```
nikhilaravi: You could also easily add a test that checks the arrow is displayed when the displayedItems array has items!

lib/hero-image-header/index.js
nelsonic: Yes, we all _understand_ the smooth scrolling _requirement_, but right now it's not traceable.
it's *always* a good idea to leave a decision trail in the form of an issue which can be referred to in the readme... simply capturing the thought process that went into selecting *this* module should not be more than a couple of lines (*unless you want to go into detail...*)

I have a *suggestion*: 
let's merge this in now, and `then` as part of debugging https://github.com/numo-labs/isearch-ui/issues/189 (*scrolling bug*) we write a few lines of docs.

nikhilaravi:
great!

PR #193, opened 2016-06-03
prevents recieve search result action from being logged

nikhilaravi:
- recieve search result action is called every time an item is sent over the socket connection - so the console is flooded with these logs and it makes it very hard to see any of the other logs!
- also added a check to disable the logger in production - before deploying we just need to set the NODE_ENV to production and the in the webpack build the logger wont be added. 
- Also updated the recieve search result reducer case to send back items to the UI until 30 have been displayed - this fixes an issue with the scroll view where the `loadMore` function wasn't being called if the user scrolls too fast to the bottom of the view and enough items haven't been received over the socket channel.


jackcarlisle:
👍  nice

PR #194, opened 2016-06-03
Updates article component to use dangerouslysetInnerHTML to render text

nikhilaravi:


src/components/article/index.js
```diff
@@ -95,7 +99,7 @@ class ArticleFullPage extends Component {
                     {key !== 0 && section.image ? <div className='articleImage' style={{backgroundImage: `url(${section.image})`}}><img
                         src={section.image}/></div> : null}
                     {section.title ? (key === 0 ? <h1>{section.title}</h1> : <h2 >{section.title}</h2>) : null}
-                    {section.text ? <p className='articleText'>{section.text}</p> : null}
+                    {section.text ? <div className='articleText' dangerouslySetInnerHTML={this.rawMarkup(section.text)}/> : null}
```
nelsonic: this method always cracks me up! 
https://facebook.github.io/react/tips/dangerously-set-inner-html.html

src/components/article/style.css
```diff
@@ -80,7 +80,8 @@ h1 {
 
 .articleFullPageContainer {
     text-align: center;
-    font-family: SourceSansLight;
+    /* commenting this out as it prevents the bold/italics etc styling on the article text being applied */
+    /*font-family: SourceSansLight;*/
```
nelsonic: @nikhilaravi is that because the *font* does not have a bold/italic option...?

src/components/isearch/index.js
```diff
@@ -64,7 +64,7 @@ class ISearch extends Component {
   }
 
   render () {
-    console.log('----RENDERING----');
+    console.log('----RENDERING----', this.props.displayedItems);
```
nelsonic: We *must* add a `Technical Task` to the `isearch-ui` backlog (*issues list*) to go through the code and remove all the `console.log` *before* we launch this next week. ***and/or*** create a `debug` variable we can set in the front-end to show the debugging console.logs on-demand.

nelsonic:
code looks good. 👍 
would consider adding comment *why* the `dangerouslySetInnerHTML` is *needed*... ❓ 
But otherwise happy to merge. 🚀

src/components/article/style.css
nikhilaravi: yes there are separate fonts for bold, italics etc which won't work with the wysiwyg editor output

src/components/isearch/index.js
nikhilaravi: yeah the other idea is to use a redux action instead of console.log - then when we can turn this on and off in the redux logger.

nikhilaravi:
@nelsonic I've updated the PR with comments.

nelsonic:
sweet! thanks! 🎉

PR #202, opened 2016-06-06
Tag view

jackcarlisle:
## ready for review

* adds searchbar and tag list to the tag view component


src/components/tag-view/index.js
```diff
@@ -15,13 +30,39 @@ export default class TagView extends Component {
               className='tagExitButton'
             />
           </Link>
```
nikhilaravi: can you remove the Link tag as you already have the onClick using a redux action?

PR #206, opened 2016-06-06
Fixes reset button

nikhilaravi:
Pressing reset, sets the tag to the homepage tag and then initiates a new search


jackcarlisle:
looks good 💯

src/actions/tags.js
```diff
@@ -112,10 +112,16 @@ export const clearSearchString = () => {
 * Action to reset all tags to empty
 */
 
-export const resetTags = (displayName, id) => {
```
lennym: The reason this was passed as an argument before was to re-use the single declaration from here - https://github.com/numo-labs/isearch-ui/blob/master/lib/tags/index.js#L5-L14

By having the tag metadata in the reducer as well we're maintaining the same data in two different places.

src/reducers/search.js
```diff
@@ -34,6 +34,11 @@ import union from 'lodash.union';
 import uniqBy from 'lodash.uniqby';
 
 export const initialState = {
+  defaultTag: {
+    displayName: 'Top inspiration',
+    colour: '#8EB8C4',
```
lennym: The colour is [determined by the tag type](https://github.com/numo-labs/isearch-ui/blob/master/lib/tags/index.js#L30), so doesn't need to be passed.

src/actions/tags.js
nikhilaravi: But the resetTags function is called by the socket service and by the tag component. This means that the default tag is declared in two places. The default tag should only need to be specified in one place so it has been moved to the reducer and removed as an argument.

src/reducers/search.js
nikhilaravi: noted. removing from the reducer.

src/actions/tags.js
lennym: So it is. I'd be tempted to fix that and try to get down to one single place where that's defined though, rather than letting it continue to spread.

nelsonic:
Thanks @nikhilaravi! 🎉

src/actions/tags.js
nikhilaravi: Yep its now only defined in the search reducer as the `defaultTag`! :)

lennym: It was also still defined in `lib/tags/index.js`

I've raised #217 to address that.

nikhilaravi: ah yes we need to remove that.  But the action doesn't use any of the arguments so it doesn't matter (https://github.com/numo-labs/isearch-ui/blob/master/src/actions/tags.js#L114) but we should definitely remove it from the file!!

lennym: The id of the default tag is also used to determine the visibility of the reset button in that file, so it is still needed there. Importing from a common module sorts that out though.

PR #209, opened 2016-06-07
Render filter tiles as part of search results

lennym:
Pass filter tiles through the same `receiveSearchResults` actions as other panels to be displayed. There are no "special case" actions or reducers specifically to handle filters differently from other panels.

Remove the notion of filter `visibleState` - a filter is either in the list of items to be displayed, or it isn't. There isn't any need to maintain a separate set of state for visbility.

When clicking on `No` on a filter panel it removes it from the displayed items list as per removing any other tile.

Color is determined programmatically from the related tag type, as per the display colors of tags.


lib/filter-tile/index.js
```diff
@@ -17,12 +17,14 @@ export default class FilterTile extends Component {
     this.setState({infoBoxVisible: !this.state.infoBoxVisible});
   }
   render () {
-    const { description, color, onYesFilter, onNoFilter, filterVisible, showAddMessage } = this.props;
+    const { description, onYesFilter, onNoFilter, showAddMessage } = this.props;
     const { infoBoxVisible } = this.state;
+    const type = description.tagid.split(':')[0];
+    const color = (type === 'geo' ? '#CEC947' : '#8EB8C4');
```
lennym: This is copied from the tag component. We may want to separate it out into a common module.

src/actions/search-results.js
```diff
@@ -125,9 +125,7 @@ export function mixDataInput () {
       const items = result.graphql.items;
 
       items.forEach(item => {
-        if (item.tile) {
-          mixture.push(item);
-        } else if (item.packageOffer) {
```
lennym: There were two clauses in the conditional which performed `mixture.push(item)`, so have that as the default and handle packages separately.

src/actions/search-results.js
```diff
@@ -164,6 +167,7 @@ export function mixDataInput () {
 
 export function startSearch () {
   return (dispatch, getState) => {
+    mixer = mixDataInput();
```
lennym: Create a new mixer per search, otherwise the counts and buffer arrays are not reset and can potentially contain results from the previous search.

src/actions/tags.js
```diff
@@ -59,18 +46,10 @@ export function deleteTag (displayName) {
 */
 
 export const onYesFilter = (displayName, id) => (dispatch) => {
-  dispatch(addSingleTag(displayName, id));
-  dispatch(onFilterClick(displayName));
-  return dispatch(startSearch());
+  return dispatch(addSingleTag(displayName, id));
```
lennym: `addSingleTag` already calls `startSearch` so there's no need to call both.

src/components/search-results/index.js
```diff
@@ -171,6 +154,18 @@ class SearchResults extends Component {
...
+            <VisbilitySensor key={index} onChange={(isVisible) => this.handleVisibility(isVisible, item)}>
+              <div key={index} className='gridItem'>
+                <FilterTile
+                  onYesFilter={onYesFilter}
+                  onNoFilter={() => removeTile(item.id)}
+                  description={item.filter}
+                />
+              </div>
+            </VisbilitySensor>
+          );
```
lennym: Filters are a standalone item type rather than a subset of the `type: 'tile'` so move the component initialisation out into its own section.

lennym: ### Query:

Each component returned in each of these conditional clauses is wrapped in the same `<VisibilitySensor/>` component. Is there a way we can DRY that out?

src/components/search-results/index.js
```diff
@@ -171,6 +154,18 @@ class SearchResults extends Component {
           } else {
             return <div/>;
           }
+        } else if (item.type === 'filter') {
+          return (
+            <VisbilitySensor key={index} onChange={(isVisible) => this.handleVisibility(isVisible, item)}>
+              <div key={index} className='gridItem'>
+                <FilterTile
+                  onYesFilter={onYesFilter}
+                  onNoFilter={() => removeTile(item.id)}
```
lennym: Call regular `removeTile` method on clicking 'No' button.

lib/filter-tile/index.js
```diff
@@ -17,12 +17,14 @@ export default class FilterTile extends Component {
...
     return (
       <div>
       {
```
jruts: since the condition is removed these outer curly braces can be removed.

lennym:
WIP as writing some additional tests for search-results component

lennym:
@jruts I've added some test for this now, and it should be ready to merge.

I have another commit for the refactor of the `VisibilitySensor` components to reduce repetition, but I'll raise that separately.

lib/article-tile/index.js
```diff
@@ -2,7 +2,7 @@ import React, { Component, PropTypes } from 'react';
 
 require('./style.css');
 
-export class ArticleTile extends Component {
+export default class ArticleTile extends Component {
```
nelsonic: what does `default` do here? 
I *thought* `default` was only necessary when the function/method/class being exported was anonymous... or when you have more than one exported method/variable ...?

lennym: I'm not 100% on ES6 modules, but I was looking at https://github.com/numo-labs/isearch-ui/pull/209/files#diff-fea5791e56ba08ff0d11c5afeac3eec3L5 and trying to work out why one of the things was not like the others, and the article tile was the only tile component that did not export a default.

lennym: That link doesn't work very well - see https://github.com/numo-labs/isearch-ui/blob/0549f4370054b98f496c8cdf2fc3e43723666e18/src/components/search-results/index.js#L1-L7

lennym: Anyway, my understanding is that `import { thing } from 'modulename';` requires knowing that the module being imported has a property of `thing`, where `import thing from 'modulename';` will take the default export and assign it to the variable `thing`, and would be preferable because it does not rely on knowledge of the internals of the module to know the primary exported object.

So, if and when a module is exporting a single "thing" (or in some cases, may have a main "thing", and some subsidiary properties) then one should export a `default` - in a manner analogous to `module.exports = ...` being the primary export in commonjs modules.

Issue #200, opened 2016-06-06
UI fixes


nikhilaravi:
Destination page
* [x] add full page view for destinations (**Jack** WIP)
* [x] add map view

Tag View page
* [x] add the current tags 
* [x] add the search bar 

Grid
* [x] Check rendering of filter tiles (**Lenny**)
* [x] Add functionality to render destination tiles 
* [x] Fix jerky scrolling on masonry or find alternative (_not fixed but decided it is okay for now_)

Edit Details page 
* [x] fix styling 
* [x] make search button functional 

Routing
* [x] Clicking a deep link and then clicking back to the package and then back to the search resets it back to the initial search - is this a bug? (**Marcelo** WIP) *this is a particular case scenario of the next error*
* [x] when sharing a link for an article/package the article/package is shown temporarily and then the page reverts to the search page (**Marcelo** WIP)
* [x] Reset button in tags should clear the tags and initiate a new search with the homepage tag 

Package Tile
* [x]  check why some star ratings are not being rendered (why?: some packages have a rating of 0)
* [x]  fix rendering of description in mobile view 
* [x]  check amenities to make sure they are not being duplicated
* [x] add 'type' label to the package offer image

lennym:
I am looking at the rendering of filter tiles at the moment.

jackcarlisle:
👍  I've assigned you to the task above, check it off once your PR has been merged

nikhilaravi:
I think these are all done so closing the issue

PR #211, opened 2016-06-07
Scrolling fixes

nikhilaravi:
Fixes bug where if the user scrolls to the bottom of the page before any tiles have loaded, no tiles will display from #200.

@tomgco I think this should fix the bug you noted.


jackcarlisle:
looks good! 👍

PR #213, opened 2016-06-07
url access fix 

Kumjami:
While accessing via URL to an article or hotel, the information was shown, but after a while we were being redirected to home, and a new search started. 

With this *hack* this behavior is avoided. 

Working on implementing a elegant solution. But this can work temporarily


jackcarlisle:
nice 👍

PR #214, opened 2016-06-07
Fixing mixing of articles and items

tomgco:


tomgco:
This is still a work in progress.

src/actions/search-results.js
```diff
@@ -108,54 +109,106 @@ export function clearFeed () {
...
+
+  function stir (amount) {
+    const pkg = mixture.filter(item => !isTile(item));
+    const ts = mixture.filter(item => isTile(item));
+    packages = pkg;
+    tiles = ts;
+    shake();
+
+    let items = mixture.splice(0, amount);
+    mixture = mixture.filter(item => true);
```
lennym: What do we think this is doing? de-referencing `mixture`?

src/actions/search-results.js
```diff
@@ -108,54 +109,106 @@ export function clearFeed () {
...
+  function shake () {
+    const total = tiles.length + packages.length;
+    // randomise tiles
+    shuffle(tiles);
+    // sort by rank packages
+
+    // weave tiles and packages to mixture
+    for (let i = 1; i < total + 1; i++) {
+      if (packages.length > 0) {
+        if (i % 6) {
```
lennym: This will mean that if there are relatively few packages compared with the number of tiles then the tiles will be sparsely spread until the end when they'll be a glut. Of course, that might be intended behaviour.

I'd be tempted to replace the magic number 6 with `Math.floor(packages.length / tiles.length)` so that tiles are roughly equally spaced amongst packages, irrespective of the ratio between them.

src/actions/search-results.js
```diff
@@ -108,54 +109,106 @@ export function clearFeed () {
...
       const items = result.graphql.items;
 
+      if (timeout) {
+        clearTimeout(timeout);
+      }
+
+      timeout = setTimeout(() => {
+        const amount = mixture.length;
+        const res = stir(amount);
+        return dispatch(receiveSearchResult(res, false, false));
```
lennym: This combination of 3 lines appears in a few different places (twice exactly as it is here, and once with `amount = 5`). It's probably a candidate for factoring out.

src/actions/search-results.js
tomgco: Removing undefined and falsy values from the array as splice leaves them in place.

src/actions/search-results.js
tomgco: > This will mean that if there are relatively few packages compared with the number of tiles then the tiles will be sparsely spread until the end when they'll be a glut. Of course, that might be intended behaviour.

This has now been handled.

> I'd be tempted to replace the magic number 6 with Math.floor(packages.length / tiles.length) so that tiles are roughly equally spaced amongst packages, irrespective of the ratio between them.

Sounds like a plan.

src/actions/search-results.js
lennym: Wouldn't it need to be `.filter(item => item)` to do that? This iterator returns true for all items, so won't exclude anything.

src/actions/search-results.js
nikhilaravi: agreed - but if there are more tiles than packages (e.g. no availability) then this will fail so maybe disperse tiles into packages if there are more packages than tiles or the reverse if there are more tiles than packages.

tomgco: > Sounds like a plan.

Actually we should tackle that in a different PR

nikhilaravi: good idea! 👍

src/actions/search-results.js
tomgco: Doh!

nikhilaravi:
@tomgco cool merging! 👍 In the next PR can you add some comments about what the mixer is doing? :)

PR #217, opened 2016-06-07
Extract the default starting tag into a constant

lennym:
Rather than defining the tag data for the default state in two different places, use a single constant file and import it where we need it.


nelsonic:
@lennym code looks good. (*better to separate this out*) 👍 
please merge master into this branch + resolve any merge conflicts.

lennym:
@nelsonic Updated

nelsonic:
👍

PR #220, opened 2016-06-07
Fixes adding of homepage tag

nikhilaravi:
- homepage tag was not being added on the index route.


jackcarlisle:
nicee 💯

jackcarlisle:
nicee 💯

PR #223, opened 2016-06-07
Remove tiles from `items` as well as `displayedItems`

### LINKED ISSUE #189: Scrolling restores previously removed tiles

lennym:
The way our infinite scroll works we maintain two arrays of items, one of the items displayed (a.k.a `displayedItems`) and one of the complete item backlog (`items`).

When removing a tile by clicking the close button it was removing only from the `displayedItems` which meant that when triggering infinite scroll and fetching new items from the backlog then this still included the removed tile, causing it to be restored into the stack.

Fix this by removing the item from both arrays when triggering the `TILES_REMOVE_TILE` action.

Fixes #189

### DIFF
--- a/src/reducers/search.js
+++ b/src/reducers/search.js
@@ -215,12 +215,15 @@ export default function search (state = initialState, action) {
         items: []
       };
     case TILES_REMOVE_TILE:
-      const filteredItems = state.displayedItems.filter(item => {
+      const iterator = item => {
         return item.id !== action.id;
-      });
+      };
+      const displayed = state.displayedItems.filter(iterator);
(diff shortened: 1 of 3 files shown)


nikhilaravi:
@lennym great!! thanks for fixing this!

PR #224, opened 2016-06-07
Scroll fixes

nikhilaravi:
Adds prop to specify scroll initial page to scroll view and sets this to 6 (as 30 items loaded initially and 5 per page) so when the user scrolls to the bottom of the first page the page is not reloaded with 5 items.

### DIFF
--- a/src/reducers/search.js
+++ b/src/reducers/search.js
@@ -38,6 +38,7 @@ export const initialState = {
     displayName: 'Top inspiration',
     id: 'marketing:homepage.dk.spies'
   },
+  scrollPage: 0,
   fingerprint: '',
   bucketId: '',
   resultId: '',
@@ -93,7 +94,8 @@ export default function search (state = initialState, action) {
     case UPDATE_DISPLAYED_ITEMS:
       return {
         ...state,
-        displayedItems: action.items
+        displayedItems: action.items,
+        scrollPage: state.scrollPage + 1
       };
     case BUSY_SEARCHING:
       return {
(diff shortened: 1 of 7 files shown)


jruts:
:+1:

src/reducers/search.js
```diff
@@ -38,6 +38,7 @@ export const initialState = {
     displayName: 'Top inspiration',
     id: 'marketing:homepage.dk.spies'
   },
+  scrollPage: 6,
```
lennym: How come this starts at 6?

nikhilaravi: 30 items loaded initially and 5 per page so when the user scrolls to the bottom of the first page the page is not reloaded with 5 items.

nikhilaravi: If you can think of a better solution please let me know!

lennym: No, I was just curious because it wasn't obvious. 

It might be good to parameterise out the 30, 6 and 5 to somewhere common though. They're spread out in a few different places and have dependencies on one another which feels a bit fragile at the moment.

lennym: In fact, I'd be tempted to maybe start it at zero (since we do start with zero results), and then update it at L87 where we fetch the 30 items. Then that way at least the 30 and the 6 are close.

Or possibly even better we could set it to `Math.ceil(itemsToDisplay.length / 5)` so that the provenance is clear.

nikhilaravi: ah okay good plan. I'll update it! Thanks

PR #225, opened 2016-06-08
map added

Kumjami:
Map added to destination view.


jackcarlisle:
looks good! 👍

src/components/destination/index.js
```diff
@@ -1,4 +1,5 @@
 import React, { Component, PropTypes } from 'react';
+import { GoogleMapLoader, GoogleMap } from 'react-google-maps';
```
nelsonic: This React component has virtually no tests: 
https://github.com/tomchentw/react-google-maps
But, it *appears* to have *enough* usage/adoption ...

PR #234, opened 2016-06-08
back button now launches a new search if previously there is no tags …

Kumjami:
…to search for

While accessing to an article or hotel via URL. After clicking *back* arrow, a new search will be launched as while reaching home page directly.


nikhilaravi:
Nice! 👍

PR #237, opened 2016-06-08
Refactor out common components from search results to reduce repetition

lennym:
In the search results component, when mapping the results to tile components, the wrapping components - i.e. `<VisibilitySensor/>` and `<div className="gridItem"/>` - were being repeated in each clause of the conditional statement on the type of tile being rendered, resulting in the same code existing in 4 places. 

Instead factor out the generating code for the inner component of the tile and then wrap all the results in the common grid item components.

Additionally adds test coverage for the search results component.


nelsonic:
Great cleanup @lennym thanks for doing this! 👍

Issue #229, opened 2016-06-08
Dismiss keyboard when 'return' key is pressed on mobile


nikhilaravi:
after a user has typed into the autocomplete bar and hit 'return' the keyboard should disappear

PR #240, opened 2016-06-08
Blur input

jackcarlisle:
## ready for review

#229 

* removes the focus on input when option has been selected from a dropdown
* prevents zoom when focusing on an input


Kumjami:
Great! :+1:

PR #243, opened 2016-06-08
Fixing filters grouping together

tomgco:
Sometimes they liked to stay close to each other, in reality it looks a bit clunky. So we now remix them back in with tiles when they next get shown.


jruts:
:+1:

package.json
```diff
@@ -51,6 +51,7 @@
     "bl": "^1.1.2",
     "chai": "^3.5.0",
     "css-loader": "^0.23.1",
+    "engine.io-client": "^1.6.9",
```
lennym: Didn't this only just get removed here - https://github.com/numo-labs/isearch-ui/commit/0a93c5a1204938c9b1b38877ac3f25a5afee0e21? How come it's come back?

PR #244, opened 2016-06-09
Destination fix

jackcarlisle:
* adds routing to the destination tile so that the full page can be viewed
* adds marker to the map view

TODO: 
* location is hard coded, change this when correct data is being sent through


nelsonic:
@jackcarlisle was that `TODO` captured somewhere in the backlog/issues?

jackcarlisle:
@nelsonic I'll add an issue now

PR #254, opened 2016-06-09
Impressions

Kumjami:
Impressions restored. 
Search results items types where changed, so they didn't match anymore with our analytic condition. This is why this functionality was lost. 

Now it's restored again. And I've decouple the analytics code for impressions and added a couple of tests. 

@nelsonic @nikhilaravi @jackcarlisle Question: I've added an analytics `helper`. I really don't know where would be the proper place. I would suggest to create in `lib` a folder for all the analytics related code. But right now our front end is completely (or almost completely) based in components and all `lib` content are components. Would it be ok for you to have at `lib` something that is not a react component? Or any other suggestions?


nelsonic:
@Kumjami looks good. 👍

PR #255, opened 2016-06-09
Restore timers

lennym:


nelsonic:
@lennym thanks v. much for clarifying and restoring that! 👍

PR #259, opened 2016-06-10
Use ellipsis UTF8 char and format over lines

tomgco:


nelsonic:
Thanks Tom. http://www.fileformat.info/info/unicode/char/2026/index.htm 👍

PR #261, opened 2016-06-10
Destination tile fix

nikhilaravi:
Fixes bug in destination tile rendering invalid html (preview text was missing an opening `<div>` in some cases)

### DIFF
--- a/lib/destination-tile/index.js
+++ b/lib/destination-tile/index.js
@@ -12,9 +12,12 @@ class DestinationTile extends Component {
     const { image, title, text } = sections[0];
     let previewTextArray = text.split('').slice(0, 292);
     if (previewTextArray[previewTextArray.length - 1] === ' ') {
-      previewTextArray = previewTextArray.slice(0, previewTextArray.length - 1);
+      previewTextArray = previewTextArray.slice(0, previewTextArray.length - 1); // remove unecessary whitespace at the end of the string
     }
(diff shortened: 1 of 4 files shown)


lib/destination-tile/index.js
```diff
@@ -12,9 +12,12 @@ class DestinationTile extends Component {
     const { image, title, text } = sections[0];
     let previewTextArray = text.split('').slice(0, 292);
     if (previewTextArray[previewTextArray.length - 1] === ' ') {
-      previewTextArray = previewTextArray.slice(0, previewTextArray.length - 1);
+      previewTextArray = previewTextArray.slice(0, previewTextArray.length - 1); // remove unecessary whitespace at the end of the string
     }
-    const previewText = previewTextArray.slice(0, previewTextArray.length - 1).join('') + '...</div>';
+    if (previewTextArray.slice(0, 5).join('').indexOf('<div>') === -1) {
+      previewTextArray = ['<div>', ...previewTextArray]; // if the paragraph doesn't start with a <div> then add one so the html is valid
```
lennym: The code and the comment don't quite align here. The comment talks about *starting* with a `<div>`, but the code is checking whether it *contains* a `<div>`.

nikhilaravi: It checks it the first 5 characters contain `<div>`. It should probably be that the first 5 characters === `<div>`! I can change that.

PR #262, opened 2016-06-10
Only show related content on non-homepage's

tomgco:


src/actions/helpers.js
```diff
@@ -34,7 +37,11 @@ export function formatQuery (store) {
   const departureAirports = constructDepartureAirportQuery(departureAirport);
   const travelPeriod = constructTravelPeriodQuery(departureDate, duration);
   const query = {passengers: passengers, travelPeriod: travelPeriod, departureAirports: departureAirports, ...formattedTags};
-  query.related = true;
+  if (!isHomepage(tags)) {
+    query.related = true;
+  } else {
+    query.related = false;
+  }
```
lennym: 5 lines of if/else seems a bit excessive for a single boolean. How about...

`query.related = !isHomepage(tags);`

or 

`query.related = isHomepage(tags) ? false : true;`

nikhilaravi:
great!

PR #263, opened 2016-06-10
Tags analytics

Kumjami:
Adding or removing a tag creates an analytics event.


nelsonic:
@Kumjami do we (_need to_) have tests for the Analytics code?

nelsonic:
Merging.

Kumjami:
@nelsonic I'm testing that the structure is what we need to have, so... if for any reason we change the structure, tests will tell us
Also by doing this I'm extracting analytics code from components classes, so it's being decoupled (tests are not needed for that though)

PR #264, opened 2016-06-10
Execute search when clicking result in autosuggest dropdown

lennym:
Automatically executes a search once option state has updated. Uses callback on `setState` because `setState` is potentially async - see https://facebook.github.io/react/docs/component-api.html

Removes the need for a setTimeout call around the blur because we no longer rely on the field being focussed to submit on pressing enter since we explicitly submit anyway.


nelsonic:
Sweet. thanks @lennym 👍

PR #266, opened 2016-06-13
Implementing per environment builds

tomgco:
This replaces `npm run build` with `npm run ci:build` and `npm run
prod:build`, this also updates the gulp tasks as well.

As a note this does remove the basename from the router history component. 

This should now use a relative path.


jruts:
Looks good :+1:

Issue #265, opened 2016-06-13
UI enhancements


nikhilaravi:
In order of priority:
* [x] Fade in the images after they are fully loaded
* [x] Hotel page loading view - empty divs in shape of final page structure - fills with content as it is loaded
* [x]  lift and shadow on hover over each tile - just for desktop 
* [x] autocomplete suggestions for when there are no matching search terms (**Marcelo WIP**)
* [x] Have several hero images so you get a new one when you reload the page

jackcarlisle:
Couple of possible links for the image fade in
http://buildwithreact.com/article/fade-in-image-recipe
http://jsbin.com/velakevahi/1/embed?css,js,output

nikhilaravi:
UI for autocomplete suggestions
![errormessage-1](https://cloud.githubusercontent.com/assets/5912647/16041838/9d32bbe4-322f-11e6-99ee-5e1817094ecf.png)

PR #274, opened 2016-06-13
Fade Image

nikhilaravi:
Created component which fades in images after they have loaded. 
Implemented for package, destination and article tiles and full pages. 

Fixes an issue in #265


nelsonic:
Looks good @nikhilaravi 👍

PR #275, opened 2016-06-13
Adds react-router-scroll to control scroll behaviour between routes 

nikhilaravi:
When the user is on the hotel page, it is scrolled to the top. But scroll position is preserved for all other pages.
 
Also updates hotel page tests

Fixes an issue from #265


nelsonic:
Good plan. 👍

PR #280, opened 2016-06-14
Dropdown data persist 

jackcarlisle:
## ready for review
* adds value to the option so that data persists when a user updates their travel info


nelsonic:
👍

PR #282, opened 2016-06-14
Tile shadow

jackcarlisle:
## ready for review 
#265 
* adds box shadow to the tiles and then increase scale with hover on desktop


nikhilaravi:
@jackcarlisle is this ready to be merged. I've reviewed it and run it locally and it looks great.

PR #285, opened 2016-06-14
Clicking search button event creates an event

Kumjami:
For now, no data will be tracked. Just the event of clicking search. 

image

### DIFF
--- a/src/components/edit-details/index.js
+++ b/src/components/edit-details/index.js
@@ -15,6 +15,9 @@ import {
 import './style.css';
 export default class EditDetails extends Component {
   onSearchClick () {
+    dataLayer.push({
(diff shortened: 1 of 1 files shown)


src/components/edit-details/index.js
```diff
@@ -15,6 +15,9 @@ import {
 import './style.css';
 export default class EditDetails extends Component {
   onSearchClick () {
+    dataLayer.push({
+      event: 'travelInfoUpdate'
```
nelsonic: @Kumjami do you think we could send the passenger mix date info to dataLayer?
(*does not have to be done now... but what are your thoughts?*)

nelsonic:
@Kumjami looks good.
merging. 👍

PR #291, opened 2016-06-15
Landing analytics

Kumjami:
When a user reach isearch, only if he is coming from spies.dk we track his data.

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -11,26 +11,28 @@
     <!-- Google Tag Manager -->
     <script src="https://cdn.rawgit.com/numo-labs/spies-url-parser/master/index.js"></script>
     <script language="javascript"> // see: https://github.com/numo-labs/isearch-ui/issues/54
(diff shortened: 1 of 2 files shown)


nelsonic:
@Kumjami / @nikhilaravi do we still nee to change _both_ `src/index.html` and `src/index.template.html` ...? they _appear_ to be the same...?

Kumjami:
@nelsonic the [production webpack config](https://github.com/numo-labs/isearch-ui/blob/master/webpack.production.config.js#L50) file is using the template one. 
I did't even though about why do we have the .template one. But I think it's a good question. @nikhilaravi could probably clarify this.

nikhilaravi:
@Kumjami @nelsonic wepback adds a hash to the filename of the bundle and so the index template is used to create an index file which has a script tag referring to the correct bundle. 

For development when we run the webpack dev server we need the index.html file. 

So yes we do need both files and any scripts for analytics need to be added to both.

Kumjami:
Good to know. Thanks @nikhilaravi :ok_hand:

nelsonic:
OK. merging. 👍

PR #294, opened 2016-06-16
departureCode added

Kumjami:
`departureCode` added to the spies click analytics event.

It would require to have https://github.com/numo-labs/lambda-ne-classic-package-provider/pull/100 merged.


nelsonic:
👍

PR #295, opened 2016-06-16
Default departure date

jackcarlisle:
## ready for review
* changes the default departure date to three months into the future


nelsonic:
Fixes: https://github.com/numo-labs/lambda-ne-classic-package-provider/issues/44 from the front-end. 👍

PR #296, opened 2016-06-16
dynamic meta title

Kumjami:
Meta title changed dynamically by `hack`. 

As  it's the only meta to be changed by now, and the change is made on front end, it makes no sense to make something more complex. 
Right now the target is not to satisfy scrappers requirements or improve SEO. And for accomplish this we would need back-end rendering, maybe:  https://github.com/reactjs/redux/blob/master/docs/recipes/ServerRendering.md
https://prerender.io/
https://github.com/Robert-W/react-prerender

So, for now, i think the simpler way is to do like this. And maybe a further step could include something from backend rendering to fit SEO requirements.

### DIFF
--- a/src/components/article/index.js
+++ b/src/components/article/index.js
@@ -87,6 +87,7 @@ class ArticleFullPage extends Component {
       const introSection = articleContent.sections[0];
       const content = articleContent.sections.slice(1);
       this.addAnalyticsData();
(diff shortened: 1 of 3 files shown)


nelsonic:
@Kumjami this is a good _short term_ solution. 👍

Issue #305, opened 2016-06-22
Title obscures input field on mobile


emiledskar:
The header text is overlapping the input field when the keyboard is open which making it difficult to see what you are typing.

<img style="-webkit-user-select: none; cursor: zoom-in;" src="https://cloud.githubusercontent.com/assets/8721221/16260912/a27dffa6-3869-11e6-93c3-1323ea485266.png" width="250">

nelsonic:
@emiledskar  we *think* this is fixed...
![screenshot_20160622-173809](https://cloud.githubusercontent.com/assets/194400/16275066/52f86d04-38a0-11e6-8bbb-a9f214d7ffd6.png)
Please re-test and confirm.

Issue #304, opened 2016-06-21
default image for tiles when there isn't one available


jackcarlisle:
![screen shot 2016-06-21 at 16 02 23](https://cloud.githubusercontent.com/assets/12450298/16234590/deaa1ea6-37c9-11e6-9e60-46fa6d6b518f.png)
Do we have a default image that we can use when a tile doesn't have one?

nelsonic:
A default image would look a bit repetitive ... let's #AskRasmus ...
relates to: #302

jackcarlisle:
sorry didn't see #302

nelsonic:
@jackcarlisle we just need to get it _fixed_ tomorrow.

PR #306, opened 2016-06-22
Image placeholder

### LINKED ISSUE #302: Missing images on Articles (404)
When visiting: http://www.tcdl.io.s3-website-eu-west-1.amazonaws.com/isearch/0.18/index.html and searching for `Spanien` we see articles for Barcelona, a few of these have "broken" images (`404`)
isearch-missing-article-images-404

jackcarlisle:
## ready for review
fixes #302 #304 #305 
* fixes the layout for the mobile view on android when searchbar is clicked
* adds placeholder image for when there isn't an image available

### DIFF
--- a/lib/article-tile/index.js
+++ b/lib/article-tile/index.js
@@ -1,12 +1,14 @@
 import React, { Component, PropTypes } from 'react';
 require('./style.css');
 import FadeImage from '../fade-image';
+import defaultImage from '../../src/constants/default-image';
 export default class ArticleTile extends Component {
   render () {
     const { tile, className } = this.props;
+    const articleImageSrc = tile.sections && tile.sections[0].image || defaultImage;
     return (
       <div className={'articleContainer ' + className}>
-        <FadeImage className='articleTileImage' src={tile.sections && tile.sections[0].image} />
+        <FadeImage className='articleTileImage' src={articleImageSrc} />
         <div className='text'>
           <h3>du har læst</h3>
           <h4>{this.props.overview}</h4>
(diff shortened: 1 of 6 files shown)


lib/article-tile/index.js
```diff
@@ -1,12 +1,14 @@
 import React, { Component, PropTypes } from 'react';
 require('./style.css');
 import FadeImage from '../fade-image';
+import defaultImage from '../../src/constants/default-image';
```
nelsonic: any reason this default image is not simply uploaded to CDN (e.g: S3) _once_?
does the `import` an image add it to the `bundle.js` or does it get shipped to S3 and loaded using "isomorphic fetch" ...?

src/constants/default-image.js
```diff
@@ -0,0 +1,3 @@
+const defaultImage = 'http://www.tcdl.io.s3-website-eu-west-1.amazonaws.com/isearch/images/thomascook_fallback-image.jpg';
```
nelsonic: is there an `https` version of this so we don't get a [`mixed content warning`](https://developer.mozilla.org/en-US/docs/Web/Security/Mixed_content#Warnings_in_Web_Console) ?

nelsonic: Try this: https://isearch-ui.s3.amazonaws.com/thomascook_fallback-image.jpg

nelsonic: @jackcarlisle would you mind updating it to: https://isearch-ui.s3.amazonaws.com/thomascook_fallback-image.jpg ?

jackcarlisle: sure!

nelsonic:
thanks for making those changes. 👍

PR #313, opened 2016-06-24
Impressions timeout (analytics)

### LINKED ISSUE #309: Analytics
- [x] **Page name**
Should be outside ecommerce objects. 

- [x] **Inflated Impressions** 
There is a lot of impressions producced when the search results are being deleted because of a new search starts. 

- [x] **Dimensions** 
All ecommerce actions should have dimensions

- [x] **Tags additions metadata** 
When a tag is added we want to know from where (article page, article tile, search bar)

- [x] **HomepageViewed too often**

Kumjami:
Consecutive tiles impressions will not be tracked unless there is more than 30 secs between impressions. 

Closes #309

### DIFF
--- a/lib/analytics-helper/index.js
+++ b/lib/analytics-helper/index.js
@@ -73,17 +73,17 @@ const _getItemId = (item) => _itemIdMapping[item.type](item);
  * Adds analytics data regarding impressions only if there no a previous impression within the last 30 seconds
  * @param item
  * @param dataLayer
- * @param auxDataLayer      Map for tracking last time a search result tile has been viewed by a user.
+ * @param impressionsTimestamp      Map for tracking last time a search result tile has been viewed by a user.
  *                          The key is the item ID and the value the timestamp when the last impression was tracked
  */
-export const addAnalyticsImpression = (item, dataLayer, auxDataLayer) => {
+export const addAnalyticsImpression = (item, dataLayer, impressionsTimestamp) => {
   const itemId = _getItemId(item);
   const currentTimestamp = (new Date()).getTime();
-  // If the element doesn't exists in auxDataLayer it means it has been never displayed to the user. So analytics event
+  // If the element doesn't exists in impressionsTimestamp it means it has been never displayed to the user. So analytics event
   // will be generated.
   // if the last impression was 30 seconds ago or before a new impression analytics event will be launched.
-  if (!auxDataLayer.has(itemId) || currentTimestamp - auxDataLayer.get(itemId) > 30000) {
-    auxDataLayer.set(itemId, currentTimestamp);
+  if (!impressionsTimestamp.has(itemId) || currentTimestamp - impressionsTimestamp.get(itemId) > 30000) {
+    impressionsTimestamp.set(itemId, currentTimestamp);
     dataLayer.push(impresionDataFactory(item));
   }
 };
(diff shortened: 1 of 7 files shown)


lib/analytics-helper/index.js
```diff
@@ -52,6 +52,39 @@ export const impresionDataFactory = (item) => {
...
+ * @param item
+ * @private
+ */
+const _getItemId = (item) => _itemIdMapping[item.type](item);
+
+/**
+ * Adds analytics data regarding impressions only if there no a previous impression within the last 30 seconds
+ * @param item
+ * @param dataLayer
+ * @param auxDataLayer
```
nelsonic: is it worth _explaining_ what `auxDataLayer` is for?

src/index.html
```diff
@@ -30,10 +30,10 @@
         'initial_search_month': (URL_PARAMS.departure_date && URL_PARAMS.departure_date.indexOf('-') > -1) ? URL_PARAMS.departure_date.split('-')[ 1 ] : '',
         'initial_search_year': (URL_PARAMS.departure_date && URL_PARAMS.departure_date.indexOf('-') > -1) ? URL_PARAMS.departure_date.split('-')[ 0 ] : '2016'
       };
-      console.log(event);
       dataLayer.push(event);
     }
     console.log('Google Tag Manager Initialised');
+    auxDataLayer = {}; // this datalayer will be used to track impressions timeouts
```
nelsonic: how is this _distinct_ from `dataLayer` ?

Kumjami: I changed it to a map. Actually I think that maybe the name is not the proper one

test/components/search-results.test.js
```diff
@@ -12,12 +12,14 @@ import sinon from 'sinon';
 
 describe('Component', function () {
   global.dataLayer = [];
+  global.impressionsTimestamp = new Map();
```
nelsonic: why is a Timestamp a `new Map()` ? (_sorry if I'm being a bit slow..._)

nelsonic: I understand why you would want to use a `Map` for this feature, but given the _incomplete_ support for it on Mobile devices: http://kangax.github.io/compat-table/es6/#test-Map Babel will transpile it to a "normal" Object ... right?

Kumjami: I'm just checking, and the browser support for `map` is not so good. So... I'm going back to an object. :sweat:

Kumjami:
@nelsonic back to object, it makes no sense and it will not work as far as im creating the map on the index.html, and it's not being *compiled* by babel
So... it's just a *fancy* solution but at the end, useless. 

Sorry for this :sweat_smile:

lib/analytics-helper/test/analytics-helper.test.js
```diff
@@ -109,4 +109,34 @@ describe('Analytics helper', function () {
...
+      const dataLayer = [];
+      const impressionsTimestamp = {};
+      analyticsHelper.addAnalyticsImpression(item, dataLayer, impressionsTimestamp);
+      expect(dataLayer).to.have.length(1);
+      done();
+    });
+    it('Should not add a new impression if there is already one within the last 30 seconds', function (done) {
+      const item = fixtures.package.item;
+      const dataLayer = [];
+      const date10secsago = (new Date()).getTime() - 10000;
```
nelsonic: @Kumjami you can also write `Date.now()` to get the timestamp in milliseconds.
see: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/now 😉

nelsonic:
@Kumjami these additions are good. merging. 👍

PR #314, opened 2016-06-27
IE compatibility 

Kumjami:
Note: Only functionallity. There are some styles to fix still. 

- [x] Edge
- [x] 11 
- [x] 10  
- [ ] 9
- [ ] 8

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,6 +46,8 @@
     "babel-istanbul": "^0.6.0",
     "babel-loader": "^6.2.3",
     "babel-plugin-transform-object-rest-spread": "^6.6.5",
+    "babel-plugin-transform-runtime": "^6.9.0",
+    "babel-polyfill": "^6.9.1",
     "babel-preset-es2015": "^6.5.0",
     "babel-preset-react": "^6.5.0",
     "babel-register": "^6.6.0",
--- a/package.json
+++ b/package.json
@@ -46,7 +46,6 @@
     "babel-istanbul": "^0.6.0",
     "babel-loader": "^6.2.3",
     "babel-plugin-transform-object-rest-spread": "^6.6.5",
-    "babel-plugin-transform-runtime": "^6.9.0",
     "babel-polyfill": "^6.9.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -46,6 +46,8 @@
     "babel-istanbul": "^0.6.0",
     "babel-loader": "^6.2.3",
     "babel-plugin-transform-object-rest-spread": "^6.6.5",
+    "babel-plugin-transform-runtime": "^6.9.0",
+    "babel-polyfill": "^6.9.1",
```
nelsonic: `babel-polyfill` has no documentation ... https://www.npmjs.com/package/babel-polyfill
babel-polyfill-no-readme

I'm sure the blog post or StackOverflow answer that instructs to use `babel-polyfill` to solve the IE Compatibility issue is convincing, but I don't _understand_ what it _does_ ... 
tried reading the code on Github: https://github.com/babel/babel/tree/master/packages/babel-polyfill 
Appears to import https://github.com/zloirock/core-js
Then what?
What is the effect on the size of `bundle.js` from importing a global JS Shim?

> This _appears_ to be the _only_ way of "_fixing_" the problem, so we should _probably_ just use it, but I want to _understand_ the impact ... 😕

nelsonic:
@Kumjami are you investigating IE 8 or 9 further or can we merge this in ?

Kumjami:
@nelsonic I have to remove the `transform-runtime` package actually. 
[Polyfill](https://babeljs.io/docs/usage/polyfill/) adds some globals that are missing from ES6 like `Array.from` (the problem we had). So it adds to global prototypes some missing properties.

I would like to fix IE9, but for now maybe its enough with 10, 11 and edge. Still I would open a different PR to fix styles. But, at least, now it's working in IE.

nelsonic:
Agree. 👍

nelsonic:
Restored branch to re-run tests with v 4.3.2 of node... don't worry I will delete it again. 👍

PR #315, opened 2016-06-27
Updates gulp script to create and push tag on gulp prod deploy

### LINKED ISSUE #319: Automatically create and upload a new tag when deploying to production
Based on the version in the package.json create a git tag and push it to remote when the gulp `prod:deploy` task is called.

nikhilaravi:
From the tag, the release still needs to be created manually - as far as I am aware, releases can only be created using the Github API.

fixes #319

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,6 +136,10 @@ gulp prod:deploy
 
 The contents of the public folder will then be uploaded to the specified Amazon S3 bucket. Have a look at the 'gulpfile.js' for implementation details.
 
+### Deploying to Production
+
+When the gulp task `prod:deploy` is called, a git tag is created based on the version number in the `package.json`. If a tag for that version exists, an error is shown. Once the tag has been created, it is pushed to github and the code is deployed to s3. Rollbacks can be done more easily if all production deployments have an associated artifict in the form of a git tag. Github _releases_ still need to be created manually from the tags. 
+
 # APPENDIX
 
 ## Setting up the React Webpack Babel Project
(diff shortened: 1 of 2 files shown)


nelsonic:
@nikhilaravi might be worth adding a paragraph to the Readme for how this works or a link to an issue where this was discussed as a comment next to the code.
It _looks_ like it works but it _might_ not be _obvious_ to our future-selves... ❓

nikhilaravi:
@nelsonic sure I will do that.

nikhilaravi:
@nelsonic I've updated the readme with notes.

lennym:
It would probably be good to have an alias for the gulp tasks in the npm scripts so we didn't need to force a global dependency.

There [seems to be a `deploy` script already](https://github.com/numo-labs/isearch-ui/blob/21d88b36f6f67ecfe91f84a365bdd79c76a90dce/package.json#L22) but it points at a non-existent gulp task from what I can work out.

nelsonic:
@lennym which _global_ dependency are you referring to?
gulp is listed in [`package.json#L62`](https://github.com/numo-labs/isearch-ui/blob/a130d76987b9b5875e7dff9a2a11bad56589a443/package.json#L62) 
I haven't installed it _globally_ and the deploy (gulp) script still works... (_sorry if I'm being a bit thick..._)

nelsonic:
@nikhilaravi thanks for adding those notes. 🎉

lennym:
@nikhilaravi If you can run `gulp deploy` from the command line then *at some point in time* you will have installed it globally. Otherwise you get an error like `bash: gulp: command not found`.

lennym:
I have created #320 to remove the need for global gulp.

PR #317, opened 2016-06-27
Show trace-request-id for debugging.

tomgco:
To activate this feature you type the following on your keyboard:
with the directions to be inputted as an arrow
`up, up, down, down, left, right, left, right, b, a`

or on mobile swipe:
`up, up, down, down, left, right, left, right, tap, tap, tap`

### DIFF
--- a/package.json
+++ b/package.json
@@ -113,7 +113,8 @@
   ],
   "semistandard": {
     "ignore": [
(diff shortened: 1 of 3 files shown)


nikhilaravi:
@tomgo looks good! cool idea!

PR #320, opened 2016-06-28
Alias gulp deploy tasks to npm scripts

lennym:
This removes the need to have gulp installed globally in order to access the gulp binary from the command line.


tomgco:
Awesome!

PR #325, opened 2016-06-28
Use semver in bucket names

hassy:


nelsonic:
@hassy I _prefer_ semver, but I think the _reason_ this was not used was to not give people the idea that the beta is a 1.0.x release ... but might as well let people get used to the idea! 😉

Issue #316, opened 2016-06-27
Query parameters for predefined entry point


jruts:
We should be able to add a query parameter to the url so that we can add tags via this way.

E.g. if we would have this `url: http://inspiationalsearch.spies.dk?search=family` I would like to see the 'Family' tag as the default tag instead of `Top inspiration`.

The search query parameter should allow multiple search tags separated by a semicolon.

You can use the autocomplete query with 1 result to translate the query param into the actual tag:
```js
query {
  viewer {
    autocomplete(text: "Family", size: 1) {
      items {
        name
        label
      }
    }
  }
}
```

lennym:
Would the expected workflow for this be as follows:

* perform a query to the autocomplete with whatever search term is provided
* add the first tag that is returned
* do a search

I assume this would leave the existing functionality around reset buttons and removing the "default" tag when adding new tags? i.e. it's not a "default" tag in the same way as "Top inspiration" is, it is simply a short-cut to a pre-filled search.

jruts:
@lennym the expected workflow would be:

1) On spies.dk there are some tiles .e.g. "family holidays"
2) Someone clicks that tile and gets redirected to the inspirational search app with "Family Holiday" in the query parameters
3) When loading the isearch component we will trigger an autocomplete with 1 result for the query parameters, and those results will indeed replace the "default" tag.

So your expectations are indeed accurate.

PR #327, opened 2016-06-28
Support an automatic search tag based on query string

lennym:
If the search is loaded with a `?search=<term>` query string parameter then do an autocomplete lookup based on the search term provided, and if a result is returned then use that as the "default" search term on initial load instead of the "Top inspiration" tag.

### DIFF
--- a/package.json
+++ b/package.json
@@ -92,6 +92,7 @@
     "lodash.union": "^4.4.0",
     "lodash.uniqby": "^4.5.0",
     "moment": "^2.12.0",
+    "querystring": "^0.2.0",
     "react": "^0.14.7",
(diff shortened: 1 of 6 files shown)


lennym:
Code complete. Unit tests imminent.

lennym:
Ready to go. 

Closes #316

nelsonic:
Good idea @lennym 👍

lennym:
I can't take any credit for the idea - that was all @jruts work - #316

nelsonic:
Yeah, we all know @jruts is the _brains_ behind this whole operation... 😉 
brains

PR #331, opened 2016-06-29
Optimise page loading

tomgco:


src/constants/heroImages.js
```diff
@@ -8,7 +8,7 @@ const heroImages = [
   'http://www.tcdl.io.s3-website-eu-west-1.amazonaws.com/isearch/images/isearch_front_8.jpg',
   'http://www.tcdl.io.s3-website-eu-west-1.amazonaws.com/isearch/images/isearch_front_9.jpg',
   'http://www.tcdl.io.s3-website-eu-west-1.amazonaws.com/isearch/images/isearch_front_10.jpg',
-  'https://cloud.githubusercontent.com/assets/12450298/14560385/62833d9a-0307-11e6-8a23-a0d1793d4e72.jpg'
+  'http://www.tcdl.io.s3-website-eu-west-1.amazonaws.com/isearch/images/isearch_front_11.jpg'
```
jruts: Is there a reason why we have the same image twice in this array?

Issue #329, opened 2016-06-29
Usabilla feedback - new


emiledskar:
* [x] ~~Let user remove article tags or at least make the UI inform the user that the tag has been added, [Usabilla 1](https://app.usabilla.com/feedback/item/35b4d3f2be35db1d092373a54b5e2201e2df4277), [Usabilla 2](https://app.usabilla.com/feedback/item/51bd74f5f63780df6c08673a6988f4ed24f60465)~~
* [x] Inside article view - The whole tag should be clickable, the text label and the circle button. [See usabilla entry](https://app.usabilla.com/feedback/item/0a54cef0a8f54c292b0a178ddcf756ff85d35dc2)
* [x] Duplicate tag for New York [See usabilla entry](https://app.usabilla.com/feedback/item/1c642b9fb94f933d39b0bb9dc2bd04401dce2ce3)
* [x] When a search term has more then five hits, the most relevant hit is sometimes hidden. E.g. when searching for "Paris" there are six (or more) tags that matches the search. Amongst those tags, the most relevant one is the one that says "Paris" but it is not visible as it is sorted below the other five tags that match the search. [See usabilla entry](https://app.usabilla.com/feedback/item/6e27145feec4853d08da333af7e19eb778be0415)
* [x] In the advanced search dialog (where you specify date) date format should be danish i.e. dd/mm-yyyy. [See usabilla entry](https://app.usabilla.com/feedback/item/4ebdf64605e24dc181ae8ce4597b32ac847a6bf2)
* [x] The reset button next to the added filters should say "NULSTIL" instead of "RESET". [See usabilla entry](https://app.usabilla.com/feedback/item/9ead54d17d9e989c205c1f3cebfb77cb0e4003a8)
* [x] All filter tiles should be completely translated to danish - now some of the text is in English. Talk to Rasmus or Martin to help out with translation. [See usabilla entry](https://app.usabilla.com/feedback/item/623d24c290fae569d8ae813210bdeb52ed412fa5)
* [x] Align this with the Spies way of presenting prices (the ppp-part). #141 & [See usabilla entry](https://app.usabilla.com/feedback/item/a0fcb373b16333f56b34b229a5ddde636cc57617)
* [x] Uppdatering should be "Opdatér" in the advanced search dialog. [See usabilla entry](https://app.usabilla.com/feedback/item/f092e8c5f673942f643f2e817d1bd3aa045fd135)
[...]

PR #333, opened 2016-06-30
Update button

jackcarlisle:
## ready for review

* fixes the update button functionality for safari
* changes the text on the update button #329


lennym:
LGTM, but as a slight nit-pick, it's really helpful for "future us" to describe the things being fixed in commit messages. "fixes the update button for safari" doesn't tell us anything about what's actually being fixed.

http://www.philandstuff.com/2014/02/09/git-pickaxe.html is a really good article about the use of descriptive commit messages.

PR #335, opened 2016-06-30
Query string

jackcarlisle:
## ready for review

* adds functionality to add tags from the query string


lennym:
Is this different to #327?

PR #337, opened 2016-06-30
Scroll fix

jackcarlisle:
## ready for review

* removes the scroll when the screen is smaller than 705px in  height


jackcarlisle:
https://github.com/numo-labs/release/issues/15

PR #342, opened 2016-06-30
Adding background colour behind loading image

tomgco:
This makes the experience on mobile nicer as the image loads from a
colour associated to the image
screenshot from 2016-06-30 16-10-44

### DIFF
--- a/lib/hero-image-header/index.js
+++ b/lib/hero-image-header/index.js
@@ -21,7 +21,8 @@ export default class Header extends Component {
   }
   render () {
     const wrapperStyle = {
-      backgroundImage: `url(${heroImageUrl})`
+      backgroundImage: `url(${heroImageUrl[0]})`,
+      backgroundColor: `${heroImageUrl[1]}`
     };
     const {
       addSingleTag,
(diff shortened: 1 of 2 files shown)


nelsonic:
@tomgco did you _manually_ detect which color to use for each image?

hassy:
Nice UX touch @tomgco :)

jackcarlisle:
@tomgco does this mean we'll only be using one image _(`heroImageUrl[0]`)_ as the header?

lib/hero-image-header/index.js
```diff
@@ -21,7 +21,8 @@ export default class Header extends Component {
   }
   render () {
     const wrapperStyle = {
-      backgroundImage: `url(${heroImageUrl})`
+      backgroundImage: `url(${heroImageUrl[0]})`,
```
nelsonic: > @jackcarlisle's point regarding the hero image being _hard-coded_ ... 
My _understanding_ of this code is that the `0` index is the url and `1` is the `background-color` ...

tomgco: Correct @nelsonic

tomgco:
> @tomgco did you manually detect which color to use for each image?

Yes, manually plugged in the best colour for it, we could have some funky stuff in here later, but seeming we only had < 10 images, thought why not.

jackcarlisle:
@tomgco my bad, I misunderstood it!

PR #348, opened 2016-07-01
tracking of written strings

Kumjami:
In order to be able to understand how users search for results we are going to track input strings. 
So everytime a user stop writting during 3 secs a new event is generated.

### DIFF
--- a/lib/analytics-helper/index.js
+++ b/lib/analytics-helper/index.js
@@ -1,3 +1,5 @@
+import defaultTag from '../../src/constants/default-tag';
+
 /**
  *
  * @type {{package: (function(): {id: string, brand: string, list: string}), filter: (function(): {id: number, brand:
@@ -96,7 +98,7 @@ export const addAnalyticsImpression = (item, dataLayer, impressionsTimestamp) =>
  * @returns {{event: string, preference_input: *, added_tags: *, displayed_tags: string}}
  */
 export const analyticsAddTagObject = (displayName, currentTags, context) => {
-  const initialTagIndex = currentTags.indexOf('Top inspiration');
+  const initialTagIndex = currentTags.indexOf(defaultTag.displayName);
   if (initialTagIndex > -1) currentTags.splice(initialTagIndex, 1);
   currentTags.push(displayName);
   return {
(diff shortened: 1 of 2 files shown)


lib/analytics-helper/index.js
```diff
@@ -96,6 +96,8 @@ export const addAnalyticsImpression = (item, dataLayer, impressionsTimestamp) =>
  * @returns {{event: string, preference_input: *, added_tags: *, displayed_tags: string}}
  */
 export const analyticsAddTagObject = (displayName, currentTags, context) => {
+  const initialTagIndex = currentTags.indexOf('Top inspiration');
```
tomgco: We should get the Tag `Top Inspiration` from the constant https://github.com/numo-labs/isearch-ui/blob/master/src/constants/default-tag.js

lib/search-bar/index.js
```diff
@@ -30,11 +32,19 @@ export default class SearchBarContainer extends Component {
...
   handleInputChange (e) {
+    const input = e.target.value;
+    clearTimeout(keyPressTimeout);
+    keyPressTimeout = setTimeout(() => {
+      const analyticsObject = {
+        event: 'searchTyping',
+        searchInput: input
+      };
+      dataLayer.push(analyticsObject);
+    }, 3000);
```
tomgco: After every keypress, 3 seconds later we push an event? Maybe we could debounce on keypress and then submit the event after x amount of time?

lib/analytics-helper/index.js
lennym: Beaten to the punch!

lib/search-bar/index.js
Kumjami: It only triggers after 3 secs if there is no another keypressed within this time.

lib/analytics-helper/index.js
Kumjami: oh!, will change it, thanks

lib/search-bar/index.js
tomgco: Missed that, cool :+1:

lib/analytics-helper/index.js
```diff
@@ -96,6 +96,8 @@ export const addAnalyticsImpression = (item, dataLayer, impressionsTimestamp) =>
...
+  if (initialTagIndex > -1) currentTags.splice(initialTagIndex, 1);
```
Kumjami: yes, @tomgco already pointed on that direction

Kumjami:
@tomgco @lennym default tag readed from file

tomgco:
LGTM!

PR #354, opened 2016-07-04
Prod

jackcarlisle:
* Deploying new version with rating icon image fix


tomgco:
@jackcarlisle PR is going in the wrong direction (should be into prod, from master).

PR #357, opened 2016-07-04
Adding deployment documentation

tomgco:


Kumjami:
Would it be interesting to add a link in the readme ? Just to see it directly in the isearch-ui front page

tomgco:
> Would it be interesting to add a link in the readme ? Just to see it directly in the isearch-ui front page

I thought about it, but the readme is getting a bit big, If everyone thinks we should then I am happy to add it in.

nelsonic:
@Kumjami we _could_ add a link the readme if new people fail to find it. 
(_I suspect its visible enough..._ 😉 )

PR #371, opened 2016-07-07
Querystring Scroll

### LINKED ISSUE #370: Scroll page down onload when auto-tagging from query string
- Page URL: http://inspirationalsearch.spies.dk/isearch/prod/?search=city-break

### Expected Behaviour

User is automatically scrolled down to their search results when a search term has been defined in the query string so that they can see the searched-for tag, and the corresponding results.

### Actual Behaviour

User stays at the top of the page, so they do not know that a search has already been carried out since they cannot see the resulting tag, or any results.

jackcarlisle:
## ready for review 
fixes #370
* scrolls to search results if query parameters are set 
* scrolls to search results after a search is performed

### DIFF
--- a/src/actions/tags.js
+++ b/src/actions/tags.js
@@ -70,7 +70,7 @@ export const searchForTag = (searchString) => (dispatch) => {
       if (autocomplete && autocomplete.items && autocomplete.items.length) {
         const tag = autocomplete.items[0];
         dispatch(addSingleTag(tag.label, tag.tagid));
-        scroll.scrollTo(window.innerHeight - 56);
+        scroll.scrollTo(window.innerHeight * 0.95);
       } else {
         dispatch(resetTags());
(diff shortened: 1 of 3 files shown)


src/actions/tags.js
```diff
@@ -69,6 +70,7 @@ export const searchForTag = (searchString) => (dispatch) => {
       if (autocomplete && autocomplete.items && autocomplete.items.length) {
         const tag = autocomplete.items[0];
         dispatch(addSingleTag(tag.label, tag.tagid));
+        scroll.scrollTo(window.innerHeight - 56);
```
nelsonic: is it _always_ **56** (_pixels_) ...?

nelsonic:
@jackcarlisle looks good. please keep an eye out if there are any issues on IE. 👍

Issue #376, opened 2016-07-12
Tile overflow & margin issue


emiledskar:
To ensure a full fix for the issue being opened can you please provide:

- Tags Searched: Spanien, n/a
- Page URL: http://inspirationalsearch.spies.dk/isearch/prod/

### Screen Shot
![Screenshot](https://cloud.githubusercontent.com/assets/8721221/16758420/a86de8e2-4811-11e6-8ba0-3ed2a853c5ca.png)
![Screenshot](https://cloud.githubusercontent.com/assets/8721221/16761216/77bed6b2-481f-11e6-8587-f3e1d8db381e.png)

### Expected Behaviour
Tile should have equal margins on both sides.
Image should be contained inside tile.

### Actual Behaviour
Left side of tile has a margin but the right side does not.
Image leaks outside of tile container.

### Further Information
First screenshot is from device toolbar in google developer mode. Here the issue is that the image is overflowing the container. 
The second screenshot is on a mobile device and here the problem is that the tile has 5px margin on the left side but not on the right.

emiledskar:
This is still visible for me on resolution 360 x *

jackcarlisle:
@emiledskar will fix today

jackcarlisle:
![screen shot 2016-07-27 at 10 07 15](https://cloud.githubusercontent.com/assets/12450298/17170663/2306e520-53e5-11e6-9740-4fbba0c312a9.png)

PR #379, opened 2016-07-12
Package Image Overflow

jackcarlisle:
## ready for review

* fixes package images #376


nelsonic:
does restricting the height and making width `100%` mean the aspect ratio will be variable?

jackcarlisle:
@nelsonic only slightly because the tile width doesn't vary a great deal _(15px)_

nelsonic:
@jackcarlisle cool. ✨

PR #380, opened 2016-07-12
Analytics

Kumjami:
- Search typing event will track number of suggestions. 
- Tags added via URL have `context='url'`


nelsonic:
👍

PR #382, opened 2016-07-13
dismissed tiles tracked

Kumjami:
Dismissed tiles are tracked now. So when pressing cross on the top right corner an analytics even will be triggered.


tomgco:
@lennym can you have a look at this seeming you are skimming the codebase around this?

PR #383, opened 2016-07-13
Tile transition after hover

jackcarlisle:
## ready for review
* adds smoother transition to the grid items after hover


Kumjami:
:+1:

Issue #388, opened 2016-07-13
Travel attributes selection popover


emiledskar:
### What to do?
Create a popover according to [this](https://zpl.io/Z1Bpx5v) 

### Why?
The user will remain in the same context when they change these attributes making it easier to understand.

### Further Information
A desktop design will be added, same functionality but some minor changes in UI.
This change will be accompanied by changing location and design of the button/link to open the popover.

emiledskar:
@jackcarlisle could you make the popover close when user clicks Opdater?

jackcarlisle:
@emiledskar sure! Will be active in the next release

PR #394, opened 2016-07-14
Travel Info Popover

jackcarlisle:
## ready for review

* converts the travel info section to a popover instead of a new page 
#388


src/components/edit-details/index.js
```diff
@@ -47,83 +46,84 @@ export default class EditDetails extends Component {
       setDuration,
       departureAirport,
       duration,
-      departureDate
+      departureDate,
+      hideTravelInfo
     } = this.props;
+    console.log('PROPS------', this.props);
```
tomgco: Unless this is needed, would it be possible to remove the `console.log`

jackcarlisle: WIll do!

Issue #392, opened 2016-07-14
New flow for setting travel attributes


emiledskar:
### What to do?

* [x] New style for search input field and new action to open travel attribute settings. See [this](https://zpl.io/194hwF) & [this](https://zpl.io/2j2jtQ), prototypes [mobile](https://marvelapp.com/3b6f839), [desktop](https://marvelapp.com/27b3b3b#13817532) to see how it should work. **[Jack WiP]**
* [x] Remove icon and action "goingplaces.svg".  
* [x] Replace with button/link [mobile](https://zpl.io/1Q2Aer), [desktop](https://zpl.io/Z1vU4xs). Action for this button should be to scroll up to the top of the page and to set focus on the search input field. 

### Why?
The goingplaces.svg icon and the action is confusing for the user. As a short term solution we will replace that icon with a link that takes the user to the top of the page.

The new style for the search input field is to help users understand how and what to search for.

The new location of the action to open travel settings is to make it more obvious to the user that they might need to change this setting to get any suggestions on packages.

PR #396, opened 2016-07-15
No Hotel Error Message

jackcarlisle:
## ready for review

#392 

* adds departure date to a section beneath the search bar
* adds  error message when no hotel packages are returned


lib/search-summary/header/index.js
```diff
@@ -1,9 +1,10 @@
 import React, { Component, PropTypes } from 'react';
 import './style.css';
-const postImage = require('../../../src/assets/goingplaces.svg');
+var Scroll = require('react-scroll');
+var scroll = Scroll.animateScroll;
```
Kumjami: why `require` instead of import ?
I'd suggest to change to this:
```js
import { animateScroll as scroll } from 'react-scroll';
```
unless there is any reason I can't see

jackcarlisle: I think Nikki mentioned a reason for doing it that way but I can't remember. I can change it

lib/tags/index.js
```diff
@@ -18,11 +18,14 @@ export default class TagContainer extends Component {
...
+      resetColour,
+      resetBackgroundColour,
+      resetBorderColour
     } = this.props;
     const resetButtonStyle = {
       color: resetColour,
-      border: '1px solid' + resetColour
+      backgroundColor: resetBackgroundColour,
+      border: '1px solid' + resetBorderColour
     };
```
Kumjami: reset button is dynamic?

jackcarlisle: it is because it changes position depending on whether you're on mobile or desktop

lib/hero-image-header/index.js
```diff
@@ -56,8 +62,20 @@ export default class Header extends Component {
                 inAutoCompleteSearch={inAutoCompleteSearch}
                 clearSearchString={clearSearchString}
                 tagView={false}
+                departureDate={departureDate}
+                showTravelInfo={showTravelInfo}
               />
             }
+            { (window.innerWidth > 750) && <Tags
```
nelsonic: desktop/tablet <--> mobile browser width break point?

src/components/isearch/index.js
```diff
@@ -214,14 +201,21 @@ class ISearch extends Component {
             clearSearchString={clearSearchString}
             searchBar
             displayedItems={displayedItems}
+            departureDate={danishDepartureDate || departOnFriday(moment().add(3, 'months')).format('DD/MM-YYYY')}
+            showTravelInfo={showTravelInfo}
+            tags={tags}
+            removeTag={removeTag}
+            resetTags={resetTags}
+            resetColour={'#F39110'}
```
Kumjami: hardcoded color, why using prop?

lib/tags/index.js
Kumjami: Then I would probably add several CSS clases, and use one or another as props, but we would still have style and logic/structure decoupled.

Kumjami: Then I would probably add several clases with the specific changes, and change add one or another via props. This would let style outside structure/logic, unless there were something that changes dynamically (e.g. if we have a width = screen.width or something like this)

src/components/isearch/index.js
jackcarlisle: because the colour changes when we use it somewhere else

lib/hero-image-header/index.js
jackcarlisle: I think this is what we previously decided our break point would be

lib/tags/index.js
jackcarlisle: I think that's the way we used to have it but it had been changed. I decided to go with the quickest fix that could be refactored later

Kumjami:
I think it's ok, just taking note to refactor styles

PR #406, opened 2016-07-18
version upgraded

Kumjami:
Releasing 1.0.24 to CI


nelsonic:
@Kumjami we _also_ need to think about releasing to Prod... 😉

Kumjami:
I would like to release to CI and let @emiledskar check that everything is ok, then, Prod

PR #419, opened 2016-07-19
Reset tags to default when removing last tag

### LINKED ISSUE #412: Result content is not cleared when tag is removed.
- Tags Searched: Spanien
- Page URL: http://inspirationalsearch.spies.dk/isearch/prod/#/?_k=bp732o
- Web Browser: http://whatsmybrowser.org/b/WBAJVLR

### Screen Shot
Screenshot

### Expected Behaviour
When I remove the tag I expect the result content to disappear and the default content to come back.

### Actual Behaviour
I can remove the tag but the result content is still there

lennym:
If there is one tag in place, and the user clicks the remove button then set the tags and search back to the default state.

Remove the remove button from the "Top inspiration" tag so that this cannot be removed.

Closes #412

### DIFF
--- a/lib/tags/index.js
+++ b/lib/tags/index.js
@@ -10,8 +10,12 @@ export default class TagContainer extends Component {
     this.props.resetTags();
   }
 
+  isNonDefaultTag (tag) {
+    return tag.id !== DEFAULT_TAG.id;
+  }
+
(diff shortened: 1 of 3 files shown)


nelsonic:
Looks good. 👍

PR #431, opened 2016-07-19
Simplify code for uniqueness check of related items

lennym:
Creates an array of ids that exist in the main search result set, then filters the related results checking if their id exists within that set.

Fixes a bug where the filter would fail and throw an error if the length of the related results was longer than the main results set.


nelsonic:
👍

Issue #414, opened 2016-07-19
Error message is shown on initial homepage


tomgco:
- Tags Searched: Homepage
- Page URL: inspirationalsearch.spies.dk/isearch/1.0.24/
- Web Browser: whatsmybrowser.org/b/1CFRE7P
Google chrome, android

### Screen Shot

![screenshot from 2016-07-19 11-54-37](https://cloud.githubusercontent.com/assets/524382/16947621/002aab9c-4da8-11e6-9c70-6d74ff94477d.png)

### Expected Behaviour

We should only see the red box when we do not find any tiles or packages.

### Actual Behaviour

Red box is always shown on homepage

jackcarlisle:
<img width="1134" alt="screen shot 2016-07-20 at 13 49 10" src="https://cloud.githubusercontent.com/assets/12450298/16986851/c59a86a8-4e80-11e6-8950-9d5d4e791441.png">

Error message removed on initial page

PR #435, opened 2016-07-20
Remove initial page error message [CI]

jackcarlisle:
## ready for review
* adds conditional so as the error message isn't displayed on initial search

#414


nelsonic:
👍

Issue #434, opened 2016-07-20
Click outside popover (on overlay) should close it 


emiledskar:
### Screen Shot

![Screenshot](https://cloud.githubusercontent.com/assets/8721221/16982431/feff45ac-4e6f-11e6-91e1-076f26862a08.png)

### Information
User should be able to click outside of the popover to close it. i.e. make click on overlay close popover.

jackcarlisle:
[video](https://drive.google.com/drive/u/0/folders/0BwJ2qQvXEjayVzdOU205MDZiaGs) of fix

tomgco:
@emiledskar Can you close this when you confirm it is working as intended on ci.

tomgco:
Actually I can confirm it is working.

PR #437, opened 2016-07-20
Popover blur [CI]

jackcarlisle:
## ready for review
* hides travel info popover when area outside is clicked   
#434


nelsonic:
@jackcarlisle looks good. but merging into `master` will only push this to CI for now. is that the desired?

jackcarlisle:
@nelsonic yeah that's ok!

Issue #403, opened 2016-07-18
No "You might also like this" section


emiledskar:
To ensure a full fix for the issue being opened can you please provide:

- Tags Searched: Aktiv Ferie
- Page URL: http://inspirationalsearch.spies.dk/isearch/prod/#/?_k=4n9qrd
- Web Browser: http://whatsmybrowser.org/b/WBAJVLR

### Usabilla
[link](https://app.usabilla.com/feedback/item/72556f3bc36893c3aa465cf36b2c8c8508935a4e)

### Expected Behaviour
I expect to see the section "You might also like this" further down the page.

### Actual Behaviour
Only one article is shown and no indication that something is loading in the background.

Kumjami:
Taking a look

Kumjami:
![image](https://cloud.githubusercontent.com/assets/8338963/16986548/663113b0-4e87-11e6-9219-a044bdcc9268.png)

maybe it would be interesting to remove the "minimal"  space between results and "might like" section (I don't know why there is gap)

jackcarlisle:
@Kumjami I think the gap is there because the div needed to be a minimum height so that we could automatically scroll to the results when a search is made

Kumjami:
@jackcarlisle ok!, I just didn't know it , thanks :)

emiledskar:
@Kumjami This seems to work perfectly fine now.

Kumjami:
@emiledskar @jackcarlisle closing it, feel free to reopen if there is something to add here

PR #439, opened 2016-07-20
rerender search results if searchComplete changes

Kumjami:
Fixes "might like" scroll issue. 

Search results pane was not being rerendered when searchComplete prop changes. This is why it was a race condition


nelsonic:
@Kumjami looks good. are we only merging this into CI?
(_please add link to the issue where this issue was described..._ https://github.com/numo-labs/isearch-ui/issues/403 ?)

PR #445, opened 2016-07-21
Sort tiles on display rather than on receipt

lennym:
This removes the tile sorting algorithm from being triggered by the receipt of tiles over the websocket to being done as part of the pull of search results from the stack to be rendered. This means that we can sort based on a larger set of data and with ranking potentially available.

Each time there is a call to load more items into the feed, the section of the result stack which is currently not displayed is sorted according to its ranking data (provided by `lambda-ian-rankin`). The top `pageSize` results are then displayed (`pageSize` currently defaults to 5).

On first load of a new search results are loaded into the feed as follows:

* loads the first 10 results as soon as they are available (configurable via `initialPageSize` prop)
* if 10 results are not found within 3 seconds then show whatever is available
* if 1s passes with no results received, and 10 are not yet available then show whatever is available

The loading is also simplified by removing the notion of page number, in favour of just loading "more" results whenever they are requested. This is equivalent to the old behaviour anyway as the page number was always incremented on loading of more results.


tomgco:
Code looks fine, need to fix the tests and let someone more familiar with how the react app to confirm and merge.

nelsonic:
LGTM

PR #455, opened 2016-07-22
Suggestions improvements

Kumjami:
- now suggestions are reset if empty input
- analytics code reorganized


nelsonic:
@Kumjami looks good. 👍

PR #461, opened 2016-07-25
onfocus no suggestions fixed if not empty input

### LINKED ISSUE #453: Search box dropdown suggestion issue
### Video
[link](https://drive.google.com/open?id=0BwJ2qQvXEjayVmt5S2tTQUxfRXM)

### Expected Behaviour
I expect to see the same suggestions in dropdown before and after focus (when the same text is still in the input field)

### Actual Behaviour
I get suggestions for no result when I search for something that is not in the system. If I then toggle focus from input field the default suggestions is displayed in the dropdown.

Kumjami:
closes #453

### DIFF
--- a/lib/search-bar/index.js
+++ b/lib/search-bar/index.js
@@ -65,16 +65,6 @@ export default class SearchBarContainer extends Component {
       document.activeElement.blur();
     });
   }
-
-  onInputFocus (e) {
-    e.preventDefault();
-    const scrollY = window.scrollY || document.documentElement.scrollTop;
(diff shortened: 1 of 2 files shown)


tomgco:
@Kumjami #453 seems to be already fixed in `1.0.24` can you confirm this? Or does your branch fix something else?

Kumjami:
@tomgco it´s not already fixed. Take into consideration that the problem is that when you write something without matches, and then you remove focus on the input (so suggestions dissapear) when the focus come back to the input, the default suggestions are displayed, instead of the "not matches" suggestions

Issue #464, opened 2016-07-25
Remove animated arrow icon.


emiledskar:
Remove the animated arrow icon that points down to the content. We do not need this after we have fixed #462 .

![image](https://cloud.githubusercontent.com/assets/8721221/17101944/631a1034-5277-11e6-801e-06adfa5c6ad6.png)

Kumjami:
Closed with #467

Issue #463, opened 2016-07-25
Remove icon in front of page titel


emiledskar:
Remove the icon in front of the page title. Users think that it is clickable so it will confuse them.
![image](https://cloud.githubusercontent.com/assets/8721221/17101733/66d42648-5276-11e6-86a6-844484c4d06b.png)

Kumjami:
closed with #467

PR #467, opened 2016-07-25
Homepage header fixs

### LINKED ISSUE #462: Title image should not cover entire page
### Information
The title image (background image) should cover about 50% of the page. Take a look at the screens in Zeplin to see what I'm referring to. 

### Why?
We want the user to see that there is default content to explore so that they can get inspired without having to search for something.

### Expected Behaviour
Something like this.
image

### Actual Behaviour
image

Kumjami:
Closes #462, #463 and #464

### DIFF
--- a/test/constants/helpers/schema.json
+++ b/test/constants/helpers/schema.json
@@ -3105,7 +3105,7 @@
           "description": "Mutations for a person who uses our app",
           "fields": [
             {
-              "name": "searchResultId",
+              "name": "startSearch",
               "description": null,
               "args": [
                 {
--- a/lib/hero-image-header/index.js
+++ b/lib/hero-image-header/index.js
@@ -2,9 +2,6 @@ import React, { Component, PropTypes } from 'react';
 import SearchBar from '../search-bar';
 import heroImages from '../../src/constants/heroImages';
 import Tags from '../tags';
-const signpost = require('../../src/assets/goingplaces.svg');
-const bouncingArrow = require('../../src/assets/down-arrow.svg');
-import { animateScroll as scroll } from 'react-scroll';
 import './style.css';
 const random = Math.floor(Math.random() * heroImages.length);
 const heroImageUrl = heroImages[random];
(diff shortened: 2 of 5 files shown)


tomgco:
Next time I think it will be easier to review if we submit fixes for bugs in a PR of their own, should make it quicker to review.

test/constants/helpers/schema.json
```diff
@@ -3105,7 +3105,7 @@
           "description": "Mutations for a person who uses our app",
           "fields": [
             {
-              "name": "searchResultId",
```
tomgco: Why is this in this PR?

lennym: My guess is because `npm test` generates it, and @Kumjami has done `git add .` or something similar. I don't actually know what this schema is used for, but since it's generated then it *probably* should be added to .gitignore.

tomgco:
I think the changes for #462 should result in a crop of the images. Currently it seems that we are going from the bottom of the image causing some weird crops like: 
screenshot from 2016-07-25 15-24-30
screenshot from 2016-07-25 15-25-45

In desktop

tomgco:
Another issue related to #462 

- Web Browser: whatsmybrowser.org/b/X5Y08NO

### Screen Shot

screenshot from 2016-07-25 15-58-42

### Expected Behaviour

Search box should be above tiles

### Actual Behaviour

Tiles are overlapping the searchbox

test/constants/helpers/schema.json
Kumjami: @lennym is right, I just forget to remove it

Kumjami:
Thanks @tomgco I will review all the pictures.
Regarding the problem with the overlay, for me suggestions were on top, but maybe it´s just because some media query.

Kumjami:
Regarding homepage header images.  Maybe I would just set the background possition at center: 

@emiledskar what do you think? 

Possition center: 

image

Current: 

image


Center: 

image


Current: 

image


Center: 
image


Current: 
image

Center: 
image

Current: 
image

Center
image

Current: 
image

Center:
image

Current: 
image

Center: 
image

Current: 
image

emiledskar:
@Kumjami Yes, lets go with center for know. If we want it to look nicer, we might need to make some new image selection

jackcarlisle:
👍 looks good! merging

PR #479, opened 2016-07-27
Grid item width fix

jackcarlisle:
## ready for review
* fixes the width of the grid items on mobile view  
#376


jackcarlisle:
screen shot 2016-07-27 at 10 07 15

PR #499, opened 2016-07-29
[Ready] suggestions height limited to 3 rows on mobile 

### LINKED ISSUE #495: Limit suggestion dropdown to 3 rows on mobile
### Screen Shot
image

### Expected Behaviour
Dropdown should be limited to 3 rows on mobile to take in to consideration the small screen area and the user's keyboard.

### Actual Behaviour
The dropdown has a default of 5 rows.

Kumjami:
[waiting @emiledskar opinion]

closes #495 

**Note**: I don't really know if we can close this or maybe we just wait until @emiledskar approves that or suggest some changes.


tomgco:
We should really do this based on viewport size instead of testing the useragent.

Kumjami:
It's based on viewport now, instead of useragent, but still I think I would wait for @emiledskar opinion before declaring it as *done*

tomgco:
> It's based on viewport now, instead of useragent, but still I think I would wait for @emiledskar opinion before declaring it as done

I concur :] 👍

jruts:
:+1:

PR #500, opened 2016-07-29
Use full result set for determining display of "no hotels" warning

lennym:
At the moment only the *displayed* results are used to determine the display of the "no hotels" message which means a lot of the time it is displayed when it shouldn't be, as at the time when the "search complete" event fires, there are often only a very small subset of the results rendered.

By using the ranking data, which is available early on, we can inspect the entire result set, and show the message only when it truly applies.


jackcarlisle:
👍

PR #501, opened 2016-07-29
[ready] empty string not tracked as input text

Kumjami:
**Ready for review**

Improved the way we track input text, now when deleting all the content we do not track the empty string. The event only triggers if the value is the current one (it may change fast between typing and receiving suggestions from back-end).

 Note: added `focusFirstSuggestions` to autosuggest element because it seems to have been lost in a recent merge.

### DIFF
--- a/lib/search-bar/index.js
+++ b/lib/search-bar/index.js
@@ -118,14 +118,16 @@ export default class SearchBarContainer extends Component {
         multiSection: multiSection
       });
       clearTimeout(keyPressTimeout);
-      keyPressTimeout = setTimeout(() => {
-        const analyticsObject = {
-          event: 'searchTyping',
-          searchInput: value,
-          suggestions: this.props.autocompleteOptions ? this.props.autocompleteOptions.length : 0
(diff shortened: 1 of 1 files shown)


jackcarlisle:
nice looks good! 👍 merging

lib/search-bar/index.js
```diff
@@ -118,14 +118,16 @@ export default class SearchBarContainer extends Component {
...
       clearTimeout(keyPressTimeout);
-      keyPressTimeout = setTimeout(() => {
-        const analyticsObject = {
-          event: 'searchTyping',
-          searchInput: value,
-          suggestions: this.props.autocompleteOptions ? this.props.autocompleteOptions.length : 0
-        };
-        dataLayer.push(analyticsObject);
-      }, 3000);
+      if (value && this.state.value === value && value !== '') {
```
lennym: `value !== ''` is implicit from the first condition (truthy `value`) so is redundant.

Also, what is the intention of the second clause?

PR #503, opened 2016-08-01
distintion between auto displayed tiles and search results tiles

Kumjami:
Items displayed by default (without previous user action) are marked as `auto displayed` for analytics purposes. 
Items displayed due to user action are marked as `search or filter results`.


jackcarlisle:
nice 👍

PR #504, opened 2016-08-02
fix item related

Kumjami:
With this: https://github.com/numo-labs/isearch-ui/pull/503, setting the related variable when it shoudn't be there, I broke the homepage visualization. 

Changing the variable to `autoDisplayed` it's clear that has a totally different objetive.


jackcarlisle:
good catch! 👍

PR #507, opened 2016-08-02
Clean up code 

lennym:
We can be more specific and efficient than a double `indexOf` check, because for the primary search the search id will fully match the saved id.


jackcarlisle:
👍

Issue #508, opened 2016-08-03
Translate from Swedish to Danish


jackcarlisle:
To ensure a full fix for the issue being opened can you please provide:

- Tags Searched: Mallorca
- Page URL: localhost:8000
- Web Browser: http://whatsmybrowser.org/b/8M4HR2D
- Mobile make & model (if applicable): n/a

### Screen Shot
![screen shot 2016-08-03 at 08 34 00](https://cloud.githubusercontent.com/assets/12450298/17357572/586eb0ce-5955-11e6-9b42-03684b00f791.png)

### Expected Behaviour
Text should be in Danish --> Vis på kort
### Actual Behaviour
Text is in Swedish --> Visa på karta

jackcarlisle:
<img width="385" alt="screen shot 2016-08-03 at 09 37 33" src="https://cloud.githubusercontent.com/assets/12450298/17359322/f21049d8-595d-11e6-975d-19811c97fef3.png">
fix

PR #509, opened 2016-08-03
Translate destination tile

jackcarlisle:
## ready for review
* translates swedish to danish on the destination tile  
#508


lennym:
I never knew you were such a linguist @jackcarlisle !

jackcarlisle:
@lennym mais oui!

PR #512, opened 2016-08-08
Ranking check

Kumjami:
Avoid display the `no hotel` message if there is no ranking info

### DIFF
--- a/src/components/search-results/index.js
+++ b/src/components/search-results/index.js
@@ -243,7 +243,7 @@ class SearchResults extends Component {
     const gridStyle = searchComplete && searchItems.length === 0 ? hideGridStyle : showGridStyle;
     return (
       <div className='gridContainer'>
-        {!isInitialTag && searchComplete && hotelItems.length === 0 ? noHotelsErrorMessage : ''}
+        {!isInitialTag && ranking && searchComplete && hotelItems.length === 0 ? noHotelsErrorMessage : ''}
         <div style={gridStyle}>
           <Masonry
             elementType={'div'}


src/components/search-results/index.js
```diff
@@ -243,7 +243,7 @@ class SearchResults extends Component {
     const gridStyle = searchComplete && searchItems.length === 0 ? hideGridStyle : showGridStyle;
     return (
       <div className='gridContainer'>
-        {!isInitialTag && searchComplete && hotelItems.length === 0 ? noHotelsErrorMessage : ''}
+        {!isInitialTag && ranking && searchComplete && hotelItems.length === 0 ? noHotelsErrorMessage : ''}
```
lennym: I don't think this is quite valid, as `ranking` will be an empty object (`{}`) by default, which will evaluate to `true` here. We probably need to check that the ranking object has at least one key.

I'd probably look to extract the `noHotelsErrorMessage` out into a function to avoid the conditional getting out of hand as well.

Kumjami: Actually, It would be more complex, because having no ranking doesn't mean we actually have hotels (so message should not be skipped maybe). 
I will change it a little bit. 
Thanks

lennym: That is what I was getting at when I talked about this last week. We have no way of knowing whether there no hotels available, or if there are hotels and they haven't returned yet.

nelsonic: @lennym in light of your most recent PR https://github.com/numo-labs/isearch-ui/pull/521 can we _safely_ know when there are no hotels available?

lennym: That is the hope, yes.

nelsonic:
@Kumjami are you able to finish this today? 😉

lennym:
@nelsonic I don't think there's anything to finish, this was superseded by #521 that does the same thing in a more rigorous way.

nelsonic:
Ok. so should we _close_ the PR then...?

nelsonic:
tumbleweed...
Closing...

PR #513, opened 2016-08-10
Destinationslider

Kumjami:
Waiting for images at destinations. (Now images are hardcoded)


Kumjami:
Now that destinations has images field, only if there is any image there it will be displayed on the header of destinations at mobile view. If there is more than one, a slider will be displayed for all the pictures.

jackcarlisle:
👍 looks good!

PR #515, opened 2016-08-10
Query params

Kumjami:
Url query params used as default search values. 

See https://github.com/numo-labs/release/issues/21

Mapping directly from https://github.com/numo-labs/isearch-ui/blob/master/src/components/edit-details/options.js

So, the only valid to display values for `travelDuration` would be 1, 2, 3 or 4. (So as to be able to display it correctly in the popup).
With travel departure code a mapping is done. So valid values are: `CPH`, `BLL`, `AAL`, `ODE` and `RNN`.

### DIFF
--- a/src/components/isearch/index.js
+++ b/src/components/isearch/index.js
@@ -51,7 +51,11 @@ class ISearch extends Component {
     });
     if (parsedQuery.travelAdults > 0) setNumberOfAdults(parsedQuery.travelAdults);
     if (parsedQuery.travelChildren > 0) setNumberOfChildren(parsedQuery.travelChildren);
-    if (parsedQuery.travelDuration > 0) setDuration(parsedQuery.travelDuration + ' uge' + (parsedQuery.travelDuration > 1 ? 'r' : ''));
+    if (!isNaN(parsedQuery.travelDuration) && parsedQuery.travelDuration > 0) {
+      // We get duration as number of days, but we are using weeks so we need to convert it.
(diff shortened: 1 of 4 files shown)


src/components/isearch/index.js
```diff
@@ -20,11 +20,40 @@ class ISearch extends Component {
...
+  }
+  loadQueryParams (params) {
+    if (!params.length) return;
+    const departureOptions = {
+      BLL: 'Billund - BLL',
+      CPH: 'København - CPH',
+      ALL: 'Aalborg - AAL',
+      ODE: 'Odense - ODE',
+      RNN: 'Rønne - RNN'
+    };
```
lennym: This feels like something that should be extracted out into constants rather than being hidden in the code for a component.

lennym:
It might be good if this had some unit tests.

Kumjami:
Adding tests

src/components/isearch/index.js
Kumjami: I will extract it. :+1:

lennym:
It's mostly ok, but there's a lot of implicit type coercion between strings and integers going on, which makes me feel uneasy.

Also, storing the duration data as a string in the form of `<n> uge(r)` only to have to [parse it back again when we make the query](https://github.com/numo-labs/isearch-ui/blob/7674f3b0c0083bf48a78746d5570463bd732e03d/src/actions/helpers.js#L98) seems a little odd. We should probably look to store the data in its raw form and format it for display at display time.

That's not really an issue for this PR in particular though, as much as the codebase in general.

PR #517, opened 2016-08-12
travelCountryCode in the URL query launches a search to get a geo tag

Kumjami:
`travelCountryCode` param is no longer ignored. 

A search will be launch to find content related with the country code while coming from spies.dk.

### DIFF
--- a/src/components/isearch/index.js
+++ b/src/components/isearch/index.js
@@ -37,7 +37,8 @@ class ISearch extends Component {
       setNumberOfAdults,
       setDepartureAirport,
       setDuration,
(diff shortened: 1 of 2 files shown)


lennym:
There's a couple of issue with this:

* the component won't wait for the websocket to connect, and so may (although unlikely) initialise a search before we have a connection id to be able to send results to.
* we're doing basically the same thing in two different places with this as the same thing with the `search` parameter.

I'd much prefer to have this bound to the websocket connection, and together with the equivalent code for the `search` parameter.

PR #518, opened 2016-08-12
Revert "travelCountryCode in the URL query launches a search to get a geo tag"

jruts:
Reverts numo-labs/isearch-ui#517


jruts:
I commited wrong PR so reverting

PR #519, opened 2016-08-12
Verson 1.2.0

Kumjami:
Like this


lennym:
I have a couple more changes to go in, so might hold off this for a short while.

PR #521, opened 2016-08-16
Track providers to determine search completeness

lennym:
Instead of a single parameter sent by the classic package provider to determine when a search is complete, instead track the completeness of all providers which respond to a search, and mark the search as complete only when all providers have returned.

This means we can reliably display/not display the "no hotels" message since we know when we are no longer expecting packages to return.


nelsonic:
👀

Issue #530, opened 2016-08-22
Google map zoom 


jackcarlisle:
### Expected Behaviour
The map automatically zooms when scrolled over
### Actual Behaviour
Zoom should be disabled when scrolling over the map to prevent users getting "_stuck_" inside the map view

PR #531, opened 2016-08-22
disable map scroll zoom 

jackcarlisle:
## ready for review

* disables scrollwheel zoom on the google map component  
#530

### DIFF
--- a/src/components/destination/index.js
+++ b/src/components/destination/index.js
@@ -41,6 +41,7 @@ class DestinationFullPage extends Component {
                 ref={(map) => (map) => console.log(map)}
                 defaultZoom={6}
                 defaultCenter={marker.position}
+                options={{scrollwheel: false}}
               >
               <Marker
                 {...marker}


jruts:
What a huge commit :fearful:

src/components/destination/index.js
```diff
@@ -41,6 +41,7 @@ class DestinationFullPage extends Component {
                 ref={(map) => (map) => console.log(map)}
```
lennym: Unrelated to this PR, but worth noting... Do we want this in prod?

nelsonic:
@jackcarlisle (I know it's a pain but) please raise an issue in JIRA before diving into fixing things so that we can show output to the decision makers... thanks! ❤️

src/components/destination/index.js
nelsonic: Probably not...

jackcarlisle:
@nelsonic sorry! been in the habit of using github issues, will raise one now

Issue #540, opened 2016-08-24
Improve Test Coverage


jackcarlisle:
## expected behaviour
Front-end code should be fully tested

## actual behaviour 
Front-end code is lacking test coverage.

PR #541, opened 2016-08-24
Improve tests

jackcarlisle:
## ready for review
- adds more front end tests to improve coverage
#540


nelsonic:
@jackcarlisle looking good so far. keep going. we ❤️  tests!

nelsonic:
@jackcarlisle have we lost our CodeCov test coverage report? is it still configured on CodeShip?

lennym:
@nelsonic Istanbul reporting was lost when we shifted the ui tests to run in karma/phantom (in order to get around the issues they had with browser APIs being required to run). 

The combination of compiling with babel and webpack and istanbul wasn't working and I had sunk days into trying to get it to work before giving up.

You're welcome to try to re-implement istanbul if you like.

nelsonic:
@lennym oh. is there an _alternative_ way of tracking/reporting the "completeness" of our tests with the karma/phantom approach...? (_apologies, I'm haven't used Karma since angular days... thanks!_)

lib/article-tile/test/index.test.js
```diff
@@ -43,10 +43,18 @@ describe('Component', function () {
...
-      expect(childrenTile).to.have.length(3);
+      expect(children).to.have.length(3);
+      done();
+    });
+    it('should render the correct child components', function (done) {
+      expect(wrapper.find('div')).to.have.length(5);
+      expect(wrapper.find('FadeImage')).to.have.length(1);
+      expect(wrapper.find('h2')).to.have.length(1);
+      expect(wrapper.find('h3')).to.have.length(1);
+      expect(wrapper.find('h4')).to.have.length(1);
```
lennym: These tests feel a bit arbitrary. 

* Containing exactly 5 `<div/>`s isn't really something we care about in the slightest, and seems like it would be very fragile in future code changes, so would just add to maintenance cost without really adding any value in terms of confirming correct behaviour.
* If we're going to assert on the headers being present, then I'd probably want to also assert that they contain the text we expect.

lib/beta-flag/test/index.test.js
```diff
@@ -0,0 +1,18 @@
...
+describe('Component', function () {
+  const wrapper = shallow(<BetaFlag />);
+  const children = wrapper.children().nodes;
+  it('should render our <BetaFlag /> component', function (done) {
+    expect(children).to.have.length(1);
+    done();
+  });
+  it('should render the correct child components', function (done) {
+    expect(wrapper.find('div')).to.have.length(1);
+    expect(wrapper.find('p')).to.have.length(1);
```
lennym: Same comments as above really, it's hard to see what the value of this test is in terms of confirming that things are working as we expect.

lib/fade-image/test/index.test.js
```diff
@@ -0,0 +1,29 @@
...
+
+const props = {
+  isBackground: true
+};
+
+describe('Component', function () {
+  it('should render our <FadeImage /> component', function (done) {
+    const wrapper = shallow(<FadeImage {...props} />);
+    const children = wrapper.children().nodes;
+    expect(children).to.have.length(0);
```
lennym: It's not obvious how this assertion relates to the functionality under test - namely `it should render our <FadeImage /> component`

lennym:
Having reviewed a good proportion of the code, there's a bit of a recurring theme of "magic number" tests, which are performing assertions on things that are not really relevant or appropriate to the actual functionality we're testing.

A particularly striking example is the following:

`expect(wrapper.find('div')).to.have.length(47);` - [here](https://github.com/numo-labs/isearch-ui/pull/541/files#diff-26efa9753f12d7990d643a6425a116fbR87)

These don't really provide any value in terms of confirming the correct functionality of components, but will add maintenance overhead when it comes to making updates to components since the tests are incredibly sensitive to change.

I'd probably want to try to be more specific about what aspects of components are key to their functionality, which generally will mean making more assertions about components in terms of the classNames and text content, and ensuring the scope of the assertions is limited only to the aspects under tests.

jackcarlisle:
@lennym thanks for the comments! Looking into changing this now!

jackcarlisle:
TODO: refactor tests for `src/components`

nelsonic:
@jackcarlisle can we merge this PR in? (_hope your "day off" is going well..._)

jackcarlisle:
@nelsonic Yeah it's good to go!

PR #542, opened 2016-08-24
Weather tile

lennym:
This is sufficient to display a very basic weather/climate tile which shows the average temperature for a destination during the month selected for departure.

The work to show the expanded data when clicking on the tile is ongoing, but it may be that we want to merge this as it is.

Example:

screen shot 2016-08-24 at 16 53 27

### DIFF
--- a/lib/analytics-helper/index.js
+++ b/lib/analytics-helper/index.js
@@ -24,6 +24,13 @@ const impressionObject = {
       'list': 'inspirational search feed'
     };
   },
+  weather: (item) => {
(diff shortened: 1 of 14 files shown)


jackcarlisle:
@lennym looks good! 👍

lennym:
Note: this will need https://github.com/numo-labs/lambda-ne-weather-provider/pull/3 to be merged and promoted to prod to work on non-CI environments.

There is currently the dev version of the weather provider on CI.

PR #546, opened 2016-08-26
Disable Map Scroll Mobile

jackcarlisle:
## ready for review
- removes the draggable property from mobile maps  
  #530

We need to find a way to test the map options of the google map so that we can see if the zoom changes once a user has scrolled onto the map

[link to issue](https://github.com/numo-labs/isearch-ui/issues/548)

### DIFF
--- a/src/components/destination/index.js
+++ b/src/components/destination/index.js
@@ -15,12 +15,17 @@ class DestinationFullPage extends Component {
     } = this.props;
     const latitude = articleContent.location ? Number(articleContent.location.lat) : 0;
     const longitude = articleContent.location ? Number(articleContent.location.lon) : 0;
+    const isDraggable = window.innerWidth > 750;
     const marker = {
       position: {
         lat: latitude,
         lng: longitude
       },
       defaultAnimation: 4
+    };
+    const mapOptions = {
+      scrollwheel: false,
+      draggable: isDraggable
     };
     return (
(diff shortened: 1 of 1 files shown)


nelsonic:
@jackcarlisle do we have a way of (*automatically*) _testing_ this?

src/components/destination/index.js
```diff
@@ -15,13 +15,18 @@ class DestinationFullPage extends Component {
     } = this.props;
     const latitude = articleContent.location ? Number(articleContent.location.lat) : 0;
     const longitude = articleContent.location ? Number(articleContent.location.lon) : 0;
+    const isDraggable = window.innerWidth > 750;
```
lennym: Is there a chance that this will still match tablets? Do we possibly want to consider a more "touch" oriented way of detecting this such as https://ctrlq.org/code/19616-detect-touch-screen-javascript?

jackcarlisle:
@lennym updated!

jackcarlisle:
@nelsonic I'll take a look at testing it

jackcarlisle:
TODO: find a way to test the map options of the google map so that we can see if the zoom changes once a user has scrolled onto the map

nelsonic:
Issue created for investigating testing Google Maps: https://github.com/numo-labs/isearch-ui/issues/548
merging for now as this feature is required.

